Issue #9, opened 2020-08-29
Prevent multiple restores from happening in parallel


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- on `2.5.0-alpha1`
- Deploy custom charts for enabling backup MCM
- Deploy resourceset and create a backup
- Currently if a restore fails and is stuck in "In Progress" state, user is able create another restore CR.

**Expected Result:**
User must be prevented from creating multiple restore CRs

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.5.0-alpha1
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): HA rke cluster
- Kubernetes version (use `kubectl version`):

```
1.18
```

mrajashree:
We don't need to prevent this. The bug causing parallel restores to fail is fixed here https://github.com/rancher/backup-restore-operator/issues/5

sowmyav27:
**On master-head - commit id: `27000be7e`**

- Deploy backup restore app - with incorrect s3 details (bucketname)
- s3 - has some backup files - file1, file2
- Deploy a couple of  restore CRs --> to restore from file1 and file2
- Restore CRs will be in Error state because S3 details are incorrect in the default storage target 
- Upgrade the backup restore app and give in the right details of S3 bucket name.
- 2 restores happen in parallel
- Logs from restore-backup-operator:

```
INFO[2020/09/26 23:19:37] Starting to restore clusterscoped resources for restore CR restore-tv6v8 
INFO[2020/09/26 23:19:37] Checking ownerRefs for resource cattle-globalrolebinding-grb-txdk4 of type rbac.authorization.k8s.io/v1, Resource=clusterrolebindings 
INFO[2020/09/26 23:19:37] Getting GVR for ownerRef management.cattle.io/v3, Kind=GlobalRoleBinding of resource cattle-globalrolebinding-grb-txdk4 
INFO[2020/09/26 23:19:37] Done waiting for CRD clusterregistrationtokens.management.cattle.io to become available 
INFO[2020/09/26 23:19:37] Starting to restore clusterscoped resources for restore CR restore-sx6f9 
```

mrajashree:
[...]

PR #41, opened 2020-09-28
Prevent multiple restores

mrajashree:
https://github.com/rancher/backup-restore-operator/issues/9

### DIFF
--- a/pkg/controllers/restore/controller.go
+++ b/pkg/controllers/restore/controller.go
@@ -18,6 +18,7 @@ import (
 	"github.com/rancher/wrangler/pkg/genericcondition"
 	"github.com/sirupsen/logrus"
 
+	coordinationv1 "k8s.io/api/coordination/v1"
 	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -29,7 +30,9 @@ import (
 	"k8s.io/apiserver/pkg/storage/value"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
+	coordinationclientv1 "k8s.io/client-go/kubernetes/typed/coordination/v1"
 	"k8s.io/client-go/util/retry"
+	"k8s.io/utils/pointer"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/controllers/restore/controller.go
```diff
@@ -700,3 +719,72 @@ func (h *handler) setReconcilingCondition(restore *v1.Restore, originalErr error
...
+	lease.Spec.HolderIdentity = leaseHolderName(restore)
+	_, err := h.kubernetesLeaseClient.Update(h.ctx, lease, k8sv1.UpdateOptions{})
+	return err
+}
+
+func (h *handler) Unlock(id string) error {
+	lease, err := h.kubernetesLeaseClient.Get(h.ctx, leaseName, k8sv1.GetOptions{})
+	if err != nil {
+		return err
+	}
```
dramich, line 776: Check the error, if it doesn't exist ignore the error.

pkg/controllers/restore/controller.go
```diff
@@ -700,3 +719,72 @@ func (h *handler) setReconcilingCondition(restore *v1.Restore, originalErr error
...
+	if lease.Spec.HolderIdentity == nil {
+		return nil
+	}
+	if *lease.Spec.HolderIdentity != id {
+		return fmt.Errorf("lease holder ID %q does not match existing restore CR", id)
```
dramich, line 781: Can this error be clarified? We are attempting to unlock and the ID does not match between the restore trying to unlock and what exists on the lease

pkg/controllers/restore/controller.go
```diff
@@ -101,6 +107,12 @@ func Register(
 		restmapper:              restmapper,
 		defaultBackupMountPath:  defaultLocalBackupLocation,
 		defaultS3BackupLocation: defaultS3,
+		kubernetesLeaseClient:   leaseClient,
+	}
+
+	lease, err := leaseClient.Get(ctx, leaseName, k8sv1.GetOptions{})
+	if err == nil && lease != nil {
+		leaseClient.Delete(ctx, leaseName, k8sv1.DeleteOptions{})
```
dramich, line 115: I think this could cause issues in the future if we have more than one running but should be fine for now.

dramich (approved):

pkg/controllers/restore/controller.go
mrajashree: made an issue for 2.5.x to track it

Issue #49, opened 2020-10-02
Mutliple scheduled backups created for hourly recurring backup using backup-restore-operator


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Add the rancher-backups app in the Cluster Explorer - Apps. Configure it to use an S3 endpoint.
2. Add a new scheduled backup with cron `45 * * * *`, retention count set to 100
3. Wait until 45 minutes after the hour.

**Result:**
Saw 13 backups. Only should see one:
![image](https://user-images.githubusercontent.com/468462/94892197-e913ca00-0438-11eb-8b51-4eeca4a7a9c8.png)
rancher-backup log:
[rancher-backup-5c5bcd7f44-5vg9c.log](https://github.com/rancher/rancher/files/5319137/rancher-backup-5c5bcd7f44-5vg9c.log)

**Other details that may be helpful:**
Checked back next day and backups were created every hour at 45 minutes passed the hour.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.5.0-rc4

- Installation option (single install/HA):
HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```
(paste the output here)
```

- Docker version (use `docker version`):

```
(paste the output here)
```

mrajashree:
This is an edge case, can be reproduced by:
1. Create a recurring backup CR, to be taken at every "Nth" minute, where N is current time+1, meaning if right now it's 12:01 PM, creating recurring backup with schedule `2 * * * *`
2. At times can be reproduced with `@every 1m`

sowmyav27:
**Reproduced this issue on 2.5-head - commit id: `ce572b5f15`**

- Create a recurring backup CR - to be taken at every "Nth" minute, where N is current time+1
- Current time then - 12:40
<img width="976" alt="Screen Shot 2020-10-02 at 12 43 50 PM" src="https://user-images.githubusercontent.com/26032343/94963397-f4a0d880-04ac-11eb-98f7-8b2977ef6ff9.png">

- Saw multiple recurring backups
[...]

PR #50, opened 2020-10-02
Don't round up nextSnapshotTS for recurring backups

mrajashree:
Problem: We calculate next snapshot TS for recurring backups and use EnqueueAfter.
We also save this nextSnapshotTS on Status. This is because at the end of processing a backup,
we update certain fields in the status, which means the handler receives another sync. We don't
want to take another backup unless current time is same as or is after NextSnapshotTS. In this
case we were rounding up nextSnapshotTS for comparison, so as to not skip another backup. Example,
if current time is 11:59:15 AM when controller receives sync, but next snapshot is at 12:00 PM, there was a chance
we could skip it, so the operator rounded up current time to compare. This leads to an edge case bug
where multiple backups will be taken until it's 12:01 PM.

Solution: Don't round up timestamp. After comparing current and next snapshot time, again calculate diff and re-enqueue the CR

https://github.com/rancher/backup-restore-operator/issues/49

### DIFF
--- a/pkg/controllers/backup/controller.go
+++ b/pkg/controllers/backup/controller.go
@@ -116,20 +116,23 @@ func (h *handler) OnBackupChange(_ string, backup *v1.Backup) (*v1.Backup, error
 			return backup, nil
 		}
 		if backup.Status.NextSnapshotAt != "" {
-			currTime := time.Now().Round(time.Minute).Format(time.RFC3339)
+			currTime := time.Now().Format(time.RFC3339)
 			logrus.Infof("Next snapshot is scheduled for: %v, current time: %v", backup.Status.NextSnapshotAt, currTime)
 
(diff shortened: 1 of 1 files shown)


pkg/controllers/backup/controller.go
```diff
@@ -116,20 +116,23 @@ func (h *handler) OnBackupChange(_ string, backup *v1.Backup) (*v1.Backup, error
 			return backup, nil
 		}
 		if backup.Status.NextSnapshotAt != "" {
-			currTime := time.Now().Round(time.Minute).Format(time.RFC3339)
+			currTime := time.Now().Format(time.RFC3339)
```
mrajashree, line 119: this isn't used anywhere, just for logging

dramich (approved):

PR #74, opened 2021-02-22
Add Fleet resources

nickgerace:
Relevant issue: https://github.com/rancher/backup-restore-operator/issues/71

### DIFF
--- a/crds/rancher-resource-set.yaml
+++ b/crds/rancher-resource-set.yaml
@@ -2,13 +2,23 @@ apiVersion: resources.cattle.io/v1
 kind: ResourceSet
 metadata:
   name: rancher-resource-set
+controllerReferences:
+  - apiVersion: "apps/v1"
+    resource: "deployments"
+    name: "rancher"
+    namespace: "cattle-system"
 resourceSelectors:
+
+  # Rancher
   - apiVersion: "v1"
     kindsRegexp: "^namespaces$"
-    resourceNameRegexp: "^cattle-|^p-|^c-|^user-|^u-|^fleet-|^cluster-fleet-"
+    resourceNameRegexp: "^cattle-|^p-|^c-|^user-|^u-"
     resourceNames:
       - "local"
-      - "rancher-operator-system"
+  - apiVersion: "apps/v1"
+    kindsRegexp: "^deployments$"
+    namespaces:
+      - "cattle-system"
   - apiVersion: "v1"
     kindsRegexp: "^secrets$"
     namespaceRegexp: "^cattle-|^p-|^c-|^local$|^user-|^u-"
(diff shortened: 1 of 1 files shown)


nickgerace:
Accidentally closed this. Reopened as draft.

crds/rancher-resource-set.yaml
```diff
@@ -29,29 +31,23 @@ resourceSelectors:
...
   - apiVersion: "v1"
     kindsRegexp: "^configmaps$"
+    namespaceRegexp: "^fleet-|^cluster-fleet-"
     namespaces:
       - "cattle-system"
-      - "fleet-system"
   - apiVersion: "rbac.authorization.k8s.io/v1"
     kindsRegexp: "^roles$|^rolebindings$"
-    namespaceRegexp: "^cattle-|^p-|^c-|^local$|^user-|^u-"
-  - apiVersion: "rbac.authorization.k8s.io/v1"
```
StrongMonkey, line 43: Why is rolebinding deleted?

dramich (commented):
Would it make sense to put the fleet things in its own section? I think it would keep the expressions from getting too big by having each main component (rancher, fleet...) separated. And future updates adding more components would be cleaner, add a new section and don't worry about existing.

paulchoi:
👍 I just migrated Rancher to another cluster, and Fleet decided to remove all managed resources from downstream clusters.
This is a very scary behavior.

nickgerace:
> Would it make sense to put the fleet things in its own section? I think it would keep the expressions from getting too big by having each main component (rancher, fleet...) separated. And future updates adding more components would be cleaner, add a new section and don't worry about existing.

@dramich had this thought while going through the PR. I think we should section it out. It's getting difficult to read.

crds/rancher-resource-set.yaml
nickgerace: This is purposeful since we moved it, but the PR is difficult to read in its current state. I'll change it. https://github.com/rancher/backup-restore-operator/pull/74#issuecomment-788536396

mrajashree:
@nickgerace @dramich  just wanted to make a note here that for the chart to pick up these changes and include them in the rancher-resource-set, the changes should be made [here](https://github.com/rancher/charts/blob/dev-v2.5-source/packages/rancher-backup/charts/templates/rancher-resourceset.yaml)
This crd file in this repo was only being maintained during the initial development/testing. 
It was also suggested (by Darren I think) to move the chart from rancher/charts repo to this repo, and then provide the chart url like its done for [rancher-operator here](https://github.com/rancher/charts/blob/dev-v2.5-source/packages/rancher-operator/package.yaml#L1)

nickgerace:
Thank you @mrajashree for the insight. That makes much more sense. We are moving the chart accordingly and I'll close this PR to avoid confusion. Thank you again and everyone for the initial reviews.

Issue #69, opened 2021-02-04
Add Fleet and Rancher Operator to the ResourceSet


sowmyav27:
From 2.5.6, fleet/rancher-operator/webhook ~will not be auto upgraded~ will not be upgraded outside of a rancher upgrade to that specific version. So this will have to be included in the resourceset - related issue - https://github.com/rancher/rancher/issues/30934

cbron:
By resourceset, do you mean the backup operator ?

nickgerace:
Dependent issues:

- [ ] Add Fleet to the ResourceSet #71
- [ ] Add Rancher Operator to the ResourceSet #72
- [ ] Add Rancher Webhook to the ResourceSet #73

nickgerace:
**Status**: #71 and #72 have been added to the chart, and the chart has been moved from `rancher/charts` to `rancher/backup-restore-operator`.

Awaiting assignment/prioritization for #73, which requires an operator-level change and was **not** included in the new chart (https://github.com/rancher/backup-restore-operator/releases/tag/v1.0.4-rc2)

nickgerace:
This issue is now only the parent issue for Fleet and Rancher Operator since Webhook requires an operator-level change.

nickgerace:
Since #72 and #71 are closed, this can be as well.

PR #79, opened 2021-03-03
[DNM 2.5.7] Add charts from rancher/charts

nickgerace:
This PR is split into **three** commits and may be easily "reviewable" by reviewing commit by commit. It follows the [rancher/rancher-operator](https://github.com/rancher/rancher-operator) charts pattern.

1. Copy chart from [rancher/charts](https://github.com/rancher/charts)
2. Add helm packaging and prepare the chart for packing

Note: we are _not_ merging the old ResourceSet file yet.

Relevant issues:
- https://github.com/rancher/tasks/issues/172
- https://github.com/rancher/backup-restore-operator/issues/78
- https://github.com/rancher/backup-restore-operator/issues/69 (while indirectly related, some resources were added to this repository's CR file, rather than the one in [rancher/charts](https://github.com/rancher/charts))

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,49 @@
 # Backup and Restore Operator
 
-
 ### Description
 
 * This operator provides ability to backup and restore Kubernetes applications (metadata) running on any cluster. It accepts a list of resources that need to be backed up for a particular application. It then gathers these resources by querying the Kubernetes API server, packages all the resources to create a tarball file and pushes it to the configured backup storage location. Since it gathers resources by quering the API server, it can back up applications from any type of Kubernetes cluster.
-* The operator preserves the ownerReferences on all resources, hence maintaining dependencies between objects. 
+* The operator preserves the ownerReferences on all resources, hence maintaining dependencies between objects.
 * It also provides encryption support, to encrypt user specified resources before saving them in the backup file. It uses the same encryption configuration that is used to enable [Kubernetes Encryption at Rest](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/). Follow the steps in [documentation link TBD] to configure this.
+
+----
+
+### Quickstart
+
+If Rancher v2.5+ is installed, you can install the `backup-restore-operator`, from the [Cluster Explorer UI](https://rancher.com/docs/rancher/v2.x/en/backups/v2.5/).
+Otherwise, you can install the charts via `helm repo` by executing the commands below.
+
+First, add our charts repository.
+
+```bash
+helm repo add rancher-charts https://raw.githubusercontent.com/rancher/charts/release-v2.5/
+helm repo update
+```
+
+Then, install both charts.
+Ensure that the CRD chart is installed first.
+
+```bash
+helm install --wait \
+    --create-namespace -n cattle-resources-system \
+    rancher-backup-crd rancher-charts/rancher-backup-crd
+helm install --wait \
+    -n cattle-resources-system \
+    rancher-backup rancher-charts/rancher-backup
+```
+
+----
+
+### Uninstallation
+
+If you are uninstalling and want to keep backup(s), ensure that you have created Backup CR(s) and that your backups are stored in a safe location.
+Execute the following commands to uninstall:
+
+```bash
+helm uninstall -n cattle-resources-system rancher-backup
+helm uninstall -n cattle-resources-system rancher-backup-crd
+kubectl delete namespace cattle-resources-system
+```
 
 ----
 
@@ -14,9 +52,9 @@
 It installs the following cluster-scoped CRDs:
 #### Backup
   A backup can be performed by creating an instance of the Backup CRD. It can be configured to perform a one-time backup, or to schedule recurring backups.
-#### Restore  
-  Creating an instance of the Restore CRD lets you restore from a backup file. 
-#### ResourceSet  
+#### Restore
+  Creating an instance of the Restore CRD lets you restore from a backup file.
+#### ResourceSet
   ResourceSet specifies the Kubernetes core resources and CRDs that need to be backed up. This chart comes with a predetermined ResourceSet to be used for backing up Rancher application
 
 ----
@@ -25,3 +63,9 @@ It installs the following cluster-scoped CRDs:
 1. Create a ResourceSet, that targets all the resources you want to backup. The ResourceSet required for backing up Rancher will be provided and installed by the chart. Refer to the default [rancher-resource-set](https://github.com/rancher/backup-restore-operator/blob/master/crds/rancher-resource-set.yaml) as an example for creating resourceSets
 2. Performing a backup: To take a backup, user has to create an instance of the Backup CRD (create a Backup CR). Each Backup CR must reference a ResourceSet. A Backup CR can be used to perform a one-time backup or recurring backups. Refer [examples](https://github.com/rancher/backup-restore-operator/tree/master/examples) folder for sample manifests
 3. Restoring from a backup: To restore from a backup, user has to create an instance of the Restore CRD (create a Restore CR). A Restore CR must contain the exact Backup filename.  Refer [examples](https://github.com/rancher/backup-restore-operator/tree/master/examples) folder for sample manifests
+
+---
+
+### Developer Documentation
+
+Refer to [DEVELOPING.md](./DEVELOPING.md) for developer tips, tricks, and workflows when working with the `backup-restore-operator`.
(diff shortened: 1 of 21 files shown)


charts/rancher-backup-crd/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+name: rancher-backup-crd
+appVersion: 999
+version: 999
```
jiaqiluo, line 4: hi @nickgerace, I am new to the development of a helm chart, please bear with me if my question is trivial: 

Any reason we bump the `apiVersion` from v1 to v2 here?  Is it just a placeholder?

jiaqiluo (changes requested):

charts/rancher-backup-crd/Chart.yaml
nickgerace: Not at problem :)
Without a `requirements.yaml`, I do not see a reason to keep `apiVersion` at `v1`. This chart is functional with Helm v3.
Reference for version bump: https://helm.sh/docs/faq/#chartyaml-apiversion-bump

jiaqiluo: Thank you for the explanation! It makes sense to me.

jiaqiluo (approved):
LGTM

charts/rancher-backup-crd/Chart.yaml
nickgerace: No problem. Actually you raise a good point that even though hand-testing proved fruitful, I want to look over this again and provide proof for this change.

nickgerace: I looked it over again and performed a backup. It works as intended, but I will leave this "unresolved" for more reviewers to take a look.

charts/rancher-backup/Chart.yaml
```diff
@@ -1,3 +1,13 @@
+apiVersion: v2
+name: rancher-backup
+appVersion: 999
+version: 999
```
rmweir, line 4: Is appVeresion and version correct here and in the other Chart.yaml?

rmweir (commented):
Why can't the CRD yaml just bee includes in `templates/` with everything else?

charts/rancher-backup/Chart.yaml
nickgerace: Yes. This is changed in `scripts/package-helm`. This reflects rancher/operator as well: https://github.com/rancher/rancher-operator/blob/master/charts/rancher-operator/Chart.yaml

nickgerace:
> Why can't the CRD yaml just bee includes in templates/ with everything else?
@rmweir (https://github.com/rancher/backup-restore-operator/pull/79#pullrequestreview-604374083)

### For "CRD yaml"

We split the CRDs into their own charts, like [rancher/rancher-operator](https://github.com/rancher/rancher-operator/tree/master/charts) and [rancher/fleet](https://github.com/rancher/fleet/tree/master/charts).
This is for customizability.

### For the CR files

In case you are referring to the `ResourceSet` CR file, that _does_ exist in `templates/`.
There is a second `ResourceSet` file in `crds/`, but that is a development artifact that is unrelated to the chart.
We are addressing that in a future PR.

README.md
```diff
@@ -9,6 +9,28 @@
 
 ----
 
+### Quickstart
+
+There are multiple ways to install the `backup-restore-operator`, including via the Cluster Explorer UI in Rancher v2.5+, but this example will focus using `helm` to install the Helm charts.
```
rmweir, line 14: focus on using*

rmweir (changes requested):

rmweir (approved):
LGTM

README.md
```diff
@@ -9,6 +9,28 @@
 
 ----
 
+### Quickstart
+
+There are multiple ways to install the `backup-restore-operator`, including via the Cluster Explorer UI in Rancher v2.5+, but this example will focus on using `helm` to install the Helm charts.
+Clone the [repository](https://github.com/rancher/backup-restore-operator), `cd` into it, and execute the following commands:
```
dramich, line 15: I think the normal way it's suggested to install charts is to add the catalog then install vs cloning a repo `helm repo add > helm install`

README.md
```diff
@@ -9,6 +9,28 @@
 
 ----
 
+### Quickstart
+
+There are multiple ways to install the `backup-restore-operator`, including via the Cluster Explorer UI in Rancher v2.5+, but this example will focus on using `helm` to install the Helm charts.
```
dramich, line 14: I think this sentence can be dropped, the operator can/is standalone

README.md
nickgerace: Should we drop this section from the `README` altogether then? The idea is to have a quick development workflow documented, but I also think it's a good idea to not give the impression that `helm install name path/to/chart` is the primary method either.

charts/rancher-backup/Chart.yaml
aiyengar2: Unresolving this comment since I'm not sure if I understand the context / agree with the idea that 999 should be used for the appVersion and version. @nickgerace can you elaborate on why this change is required?

From what I see in the scripts/package-helm script, it shouldn't matter if this is an actual version or if this is a 999, so I think we should leave the appVersion / version here and keep it up to date, even if the final chart will have its version "fixed" by the packaging script.

README.md
```diff
@@ -9,6 +9,28 @@
...
+Clone the [repository](https://github.com/rancher/backup-restore-operator), `cd` into it, and execute the following commands:
+
+```bash
+helm install --create-namespace -n cattle-resources-system rancher-backup-crd charts/rancher-backup-crd
+helm install -n cattle-resources-system rancher-backup charts/rancher-backup
+```
+
+You can continue to the [User flow](#User-flow) from here, or uninstall the charts.
+If you are uninstalling and want to keep backup(s), ensure that you have created Backup CR(s) and that your backups are stored in a safe location.
```
dramich, line 23: Make uninstall its own header and section

aiyengar2 (changes requested):
@nickgerace, once we add this to this repository, can we tie this to a PR on rancher/charts@dev-v2.5-source to treat Rancher Backup as a forked repository from now on? That way we only make updates in this repository and just update the pointer on rancher/charts, similar to what we do for fleet today.

README.md
dramich: Well, quickstart and development are two different things. This being a helm chart it needs to have proper helm install instructions, that would include adding our repo and then installing. If development docs are needed then add those to a separate section/file for contributing.

nickgerace: In that case, I'll drop this Quickstart section then and add a contributing/development file. I think that would work better in the context of this PR.

README.md
nickgerace: Will move it to the dev/contrib file.

README.md
dramich: We need the quickstart and it needs to show how to install this chart using our repo.

charts/rancher-backup/Chart.yaml
nickgerace: @aiyengar2 I think setting it to a static value could be problematic if we miss it. It's not backed by CI. We would have to "manually" ensure that the versions are correct in this `Chart.yaml`.

Moreover, I think this chart will only be manually installed for development purposes, like `rancher/rancher-operator`. Having it follow that pattern would ensure that we don't "miss a spot" when updating the chart.

With that in mind my answer is a question: if we aren't using the version strings anyway since the operator is primarily installed via the Cluster Explorer UI, then what purpose do we have in tracking these versions?

If we determine that we do need to track these versions, then perhaps we change the CI to reflect that. Thoughts?

README.md
nickgerace: I'll do both then

nickgerace:
> @nickgerace, once we add this to this repository, can we tie this to a PR on rancher/charts@dev-v2.5-source to treat Rancher Backup as a forked repository from now on? That way we only make updates in this repository and just update the pointer on rancher/charts, similar to what we do for fleet today.

@aiyengar2 already ahead of you ;) https://github.com/rancher/charts/pull/1021

charts/rancher-backup/Chart.yaml
nickgerace: Re-resolving this conversation after a conversation with @ibuildthecloud and @aiyengar2.

charts/rancher-backup-crd/Chart.yaml
nickgerace: Confirmed, we are good to go.

dramich (approved):

Issue #73, opened 2021-02-12
Add Rancher Webhook to the ResourceSet


nickgerace:
~~Parent issue: https://github.com/rancher/backup-restore-operator/issues/69~~

This issue no longer has a parent issue due to change in scope.

mrajashree:
Rancher webhook resourceSet will contain validating webhook objects, and when it was a part of the Rancher resourceSet it didn’t get restored correctly due to the current order in which we restore resources. This issue has more details on how we might have to slightly change the restore order for webhooks https://github.com/rancher/backup-restore-operator/issues/45

nickgerace:
Thank you for the context @mrajashree!

nickgerace:
Removal PR for reference: https://github.com/rancher/charts/pull/726

nickgerace:
This is dependent on restore working first: https://github.com/rancher/backup-restore-operator/issues/45

cbron:
Bumping to 2.6

jiaqiluo:
Closing this issue because we do not need to backup the rancher-webhook app, or at least the ValidatingWebhookConfiguration anymore.

See [this comment](https://github.com/rancher/backup-restore-operator/issues/45#issuecomment-834683142) for a detailed explanation.

Issue #72, opened 2021-02-12
Add Rancher Operator to the ResourceSet


nickgerace:
Parent issue: https://github.com/rancher/backup-restore-operator/issues/69

sowmyav27:
**On 2.5-head commit id: `2f2cfe6`**

- Install a HA setup on 2.5-head
- deploy two downstream clusters
- Add a git repo in continuous Delivery Tab in Cluster explorer
- Verify the apps are deployed And on continuous Delivery Tab in Cluster explorer, Clusters are seen Active
- Deploy rancher-backup `rancher-backup:1.0.400-rc02` and take a backup b1
- Followed these steps to migrate to a new Rancher setup- https://rancher.com/docs/rancher/v2.x/en/backups/v2.5/migrating-rancher/ (`rancher-backup:1.0.400-rc02` is installed)
- On the new Rancher setup, added new cluster.
- git repo in continuous Delivery Tab in Cluster explorer is deployed on the new cluster successfully.

Issue #45, opened 2020-09-29
Update the order of restore for Admission webhooks


mrajashree:
When restoring rancher-webhook was tried [here](https://github.com/rancher/backup-restore-operator/issues/44), it failed because of the order in which restore is done
1. CRDs first
2. Cluster-scoped resources
3. Namespaced resources

Because of this, the validating webhook config that is cluster-scoped gets restored, before its service is restored since the service is namespaced. So creation of any objects that go through validating webhook will fail. This can be fixed by the following restore order
1. CRDs
2. Cluster-scoped (except admission webhooks validating/mutating)
3. Namespaced
4. Validating/Mutating admission webhooks

nickgerace:
Parent issue: https://github.com/rancher/backup-restore-operator/issues/73

cbron:
Bumping to 2.6

jiaqiluo:
Closing this issue since we decided to exclude the ValidatingWebhookConfiguration resources from the backup. 

Besides the ValidatingWebhookConfiguration resource, there is a resource which type is MutatingWebhookConfiguration and never be backuped. 

Here are why we can ignoer them:
- Both the ValidatingWebhookConfiguration and MutatingWebhookConfiguration resources are created and managed by the rancher-webhook app internally.
- Rancher-webhook app is managed by Rancher server internally and does not have any user-generated data.
- No matter the rancher-webhook app or resources it manages are restored or not, Rancher server will run a helm command to either install or upgrade the app after Rancher runs up, then rancher-webhook will create/upgrade other resources. 


Theoretically, we can also exclude the rancher-webhook app from the backup, which requires a change on the regular expression used for matching deployment, secret, and service account in the namespace cattle-system. But because keeping rancher-wehook's deployment, secret, and service account does not cause any problem, we decided to keep the regular expression as it is today.

PR #80, opened 2021-03-05
Reorganize ResourceSet CR file, add Fleet and Operator

nickgerace:
This PR reorganizes the ResourceSet file, removes the development ResourceSet file, and ensures that `fleet` and `rancher-operator` are added.

**Relevant issues:**
- https://github.com/rancher/tasks/issues/172
- #71, #72
- #69 

**Commits**
1. Update old resourceset file in place
2. Delete the dev resourceset and add its contents to the chart
3. Modularize the resourceset and update README to reflect that

**Note**: the webhook in #69 (child issue #73) is not being included in this PR since that relies on #45 being resolved.


jiaqiluo:
@nickgerace  Can you update the link in the [userflow](https://github.com/rancher/backup-restore-operator#user-flow) section in the README.md file because `crds/rancher-resource-set.yaml` will be removed?

jiaqiluo:
hi @nickgerace  How is your PR address this issue: https://github.com/rancher/backup-restore-operator/issues/45?  Can you elaborate on that a little bit?

nickgerace:
> hi @nickgerace How is your PR address this issue: #45? Can you elaborate on that a little bit?

@jiaqiluo this PR _does not_ address that issue. It would have included the webhook (#73), but there is an operator-level change that should be addressed in another PR (#45). Thus, this only includes Fleet and rancher-operator, which is two-thirds of #69.

ryansann (approved):
LGTM

jiaqiluo (approved):
LGTM

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
```diff
@@ -0,0 +1,49 @@
...
+  namespaceRegexp: "^fleet-|^cluster-fleet-"
+- apiVersion: "rbac.authorization.k8s.io/v1"
+  kindsRegexp: "^roles$|^rolebindings$"
+  namespaceRegexp: "^fleet-|^cluster-fleet-"
+- apiVersion: "rbac.authorization.k8s.io/v1"
+  kindsRegexp: "^clusterrolebindings$"
+  resourceNameRegexp: "^fleet-|^gitjob-"
+- apiVersion: "rbac.authorization.k8s.io/v1"
+  kindsRegexp: "^clusterroles$"
+  resourceNameRegexp: "^fleet-"
```
aiyengar2, line 29: does this need cluster-fleet?

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
```diff
@@ -0,0 +1,49 @@
...
+  namespaceRegexp: "^fleet-|^cluster-fleet-"
+- apiVersion: "v1"
+  kindsRegexp: "^configmaps$"
+  namespaceRegexp: "^fleet-|^cluster-fleet-"
+- apiVersion: "rbac.authorization.k8s.io/v1"
+  kindsRegexp: "^roles$|^rolebindings$"
+  namespaceRegexp: "^fleet-|^cluster-fleet-"
+- apiVersion: "rbac.authorization.k8s.io/v1"
+  kindsRegexp: "^clusterrolebindings$"
+  resourceNameRegexp: "^fleet-|^gitjob-"
```
aiyengar2, line 26: Gitjob was never tracked as a regexp:

```yaml
  kindsRegexp: "^clusterrolebindings$"
  resourceNameRegexp: "^cattle-|^clusterrolebinding-|^globaladmin-user-|^grb-u-|^fleet-"
  resourceNames:
    - "eks-operator"
    - "rancher-operator"
    - "gitjob-binding"
```

What other gitjob- bindings do we want to collect? is there a reason why we collect only gitob bindings but not the namespaces, secrets, serviceaccounts, configmaps, or rolebindings?

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
```diff
@@ -0,0 +1,49 @@
...
+- apiVersion: "rbac.authorization.k8s.io/v1"
+  kindsRegexp: "^clusterroles$"
+  resourceNameRegexp: "^fleet-"
+  resourceNames:
+    - "gitjob"
```
aiyengar2, line 31: any reason why we don't collect all clusterroles with gitjob- as well?

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
```diff
@@ -0,0 +1,49 @@
+- apiVersion: "v1"
+  kindsRegexp: "^namespaces$"
+  resourceNameRegexp: "^fleet-|^cluster-fleet-"
```
aiyengar2, line 3: @nickgerace IMO, it might make sense to start each file with something like:
```yaml
{{- $resourceNameRegexp := "^fleet-|^cluster-fleet-" }}
{{- $resourceNamespaceRegexp := "^fleet-|^cluster-fleet-" }}
```

And then refer to them as:
```yaml
resourceNameRegexp: {{ $resourceNameRegexp }}
namespaceRegexp: {{ $resourceNamespaceRegexp }}
```

That way specific overrides to the regexp expressions can be called out in the template itself, such as adding something like `gitjob` for only certain components.

aiyengar2 (changes requested):
At the moment, the commit seems to be a mix between copying changes directly from `crds/rancher-resource-set.yaml` and manual changes introduced by @nickgerace.

I picked through all of these changes and documented them in my review (see all comments that start with `Original differs from current`). Perhaps it might make sense to consider splitting any changes to a different commit so we can revert them independently / call out each change we made?

Also, amongst the changes that were manually added, I only reviewed those that previously existed and were modified. @nickgerace were there any additional resourceSelectors that were added for any of the components that were untracked in the original?

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
nickgerace: They are only called "gitjob"

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
nickgerace: It is only called "gitjob-binding". As for the second part, are you asking if gitjobs create those unique resources?

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
nickgerace: No

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
aiyengar2: Yes, but more broadly I guess my question is whether that would ever be in scope for the gitjob chart or if we are sure it will never be in scope.

If so, maybe we should future-proof this chart to include pick up all resources with `gitjob` as the name or `gitjob-*` by default?

nickgerace: There's no harm in adding both, so I will.

nickgerace:
Reorganizing this PR into two commits for readability.

nickgerace:
Okay @aiyengar2 @jiaqiluo @ryansann - I split this into three commits for readability. It should be easier to review now by looking commit by commit.

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
```diff
@@ -0,0 +1,49 @@
...
+        values: ["helm"]
+      - key: "fleet.cattle.io/managed"
+        operator: "In"
+        values: ["true"]
+- apiVersion: "v1"
+  kindsRegexp: "^serviceaccounts$"
+  namespaceRegexp: "^fleet-|^cluster-fleet-"
+- apiVersion: "v1"
+  kindsRegexp: "^configmaps$"
+  namespaceRegexp: "^fleet-|^cluster-fleet-"
```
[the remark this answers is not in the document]
aiyengar2: nvm, this looks good to me. Better to collect more than less.

charts/rancher-backup/files/default-resourceset-contents/eks.yaml
```diff
@@ -0,0 +1,17 @@
+- apiVersion: "eks.cattle.io/v1"
+  kindsRegexp: "."
+- apiVersion: "apps/v1"
```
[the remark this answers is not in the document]
aiyengar2: Based on https://github.com/rancher/system-charts/blob/42664a862a58e21b1d4c2033519a44da58539333/charts/rancher-eks-operator/1.0.700/templates/deployment.yaml#L5, I think this is a reasonable assumption.

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
aiyengar2: Based on the current version of the commits, I think it's fine to omit this. But this is something probably to consider in the future.

aiyengar2 (approved):
ship it!

nickgerace:
Squashing all three commits after re-reviews after discussing with @aiyengar2.

PR #84, opened 2021-03-16
Fix CI for charts and move CRD creation to charts 

nickgerace:
Relevant issues:
- rancher/tasks#172
- #71, #72
- #69


nickgerace:
Helm install was taken from [rancher/rancher-operator](https://github.com/rancher/rancher-operator/blob/master/Dockerfile.dapper#L11)

ryansann (approved):

scripts/version
```diff
@@ -21,7 +21,11 @@ SUFFIX="-${ARCH}"
 
 TAG=${TAG:-${VERSION}${SUFFIX}}
 REPO=${REPO:-rancher}
+HELM_TAG=${TAG:-${VERSION}}
+HELM_VERSION=${HELM_TAG/v/}
 
 if echo $TAG | grep -q dirty; then
     TAG=dev
+    HELM_TAG=dev
+    HELM_VERSION=0.0.0-dev
```
jiaqiluo, line 30: Why do we need them? Where are they used?

jiaqiluo (changes requested):

scripts/version
nickgerace: We need them in scripts/package-helm. They were not present in the original PR that added scripts/package-helm

jiaqiluo (approved):
LGTM

charts/rancher-backup/values.yaml
```diff
@@ -1,6 +1,6 @@
 image:
   repository: rancher/backup-restore-operator
-  tag: v1.0.3
+  tag: latest
```
jiaqiluo, line 3: Why do we need to change this tag?  I noticed its real value will be determined by the build scripts. 

`rancher/backup-restore-operator:latest` does not exist in dockerhub.

jiaqiluo (changes requested):

charts/rancher-backup/values.yaml
nickgerace: The thought was that, like `rancher-operator`, we would use a development tag in-tree. This gets overwritten by `scripts/package-helm` anyway, like you said.

However, you are correct that `latest` does not exist, so I will revert this change for this PR. We will most likely need to ensure that it gets changed in the future.

jiaqiluo (approved):
LGTM

ryansann (approved):

PR #87, opened 2021-04-07
[DNM v2.6/v2.5.9] Restore webhooks after cluster-scoped and namespaced resources

nickgerace:
Relevant issues:
- https://github.com/rancher/backup-restore-operator/issues/45
- https://github.com/rancher/backup-restore-operator/issues/73

PR #89, opened 2021-04-12
[DNM v2.6/v2.5.9] Add Rancher Webhook to default resourceset

nickgerace:
Dependent on #87 being merged, but is ready for review.

Note to reviewers: https://github.com/rancher/backup-restore-operator/pull/89#issuecomment-817950489

Relevant issues:
- #45
- #73


nickgerace:
**Note to reviewers:** there is a service account, deployment, and service created in `cattle-system`. Since the `main.yaml` file covers all instances of those resources in that namespace, we do not need to add it to the new file.

charts/rancher-backup/files/default-resourceset-contents/main.yaml
```diff
@@ -18,6 +18,9 @@
 - apiVersion: "v1"
   kindsRegexp: "^serviceaccounts$"
   namespaceRegexp: "^cattle-|^p-|^c-|^local$|^user-|^u-"
+- apiVersion: "v1"
+  kindsRegexp: "^services$"
+  namespaceRegexp: "^cattle-|^p-|^c-|^local$|^user-|^u-"
```
jiaqiluo, line 23: I think we should select only the webhook's one by calling out its name `rancher-webhook`. 
 ```suggestion
   namespaceRegexp: "^cattle-|^p-|^c-|^local$|^user-|^u-"
   resourceNames:
     - "rancher-webhook"
```

jiaqiluo (changes requested):

charts/rancher-backup/files/default-resourceset-contents/main.yaml
nickgerace: Hmm, probably a good idea. However, we should keep deployments and serviceaccounts the same.

jiaqiluo: it is a good point.  
It is ok that we just backup all services under the target namespaces if that is how we select other resources

jiaqiluo (approved):
LGTM

ryansann (approved):

nickgerace:
Closing since https://github.com/rancher/backup-restore-operator/issues/73 has been closed

Issue #95, opened 2021-04-23
Backup clusterRoleBinding which name has  the prefix  `crb-`


jiaqiluo:
Part of the fix for this [issue](https://github.com/rancher/rancher/issues/29932) is to introduce two new naming patterns:

- new clusterRoleBinding will be named [crb-*](https://github.com/rancher/rancher/blob/1c3629fbe520d4cfe66b95c31cfdd2f3946f39dc/pkg/rbac/common.go#L179) 
- new roleBinding will be named [rb-*](https://github.com/rancher/rancher/blob/1c3629fbe520d4cfe66b95c31cfdd2f3946f39dc/pkg/rbac/common.go#L169)  

We need to add the new clusterRoleBinding to the resourceSet.
The new roleBinding is already covered.

sowmyav27:
**Verified using `1.0.400-rc04`**

- clusterrolebindings starting with `crb-` resource names are also backed up.
- P1 usecase of migrating rancher on 2.5.8-rc9 works fine. 
- Had a cluster, assigned PO/PM and CO.

PR #96, opened 2021-04-23
add new name regexp for clusterrolebinding for Rancher

jiaqiluo:
issue: https://github.com/rancher/backup-restore-operator/issues/95

New clusterRoleBinding will be named [crb-*](https://github.com/rancher/rancher/blob/1c3629fbe520d4cfe66b95c31cfdd2f3946f39dc/pkg/rbac/common.go#L179) 
New roleBinding will be named [rb-*](https://github.com/rancher/rancher/blob/1c3629fbe520d4cfe66b95c31cfdd2f3946f39dc/pkg/rbac/common.go#L169)  

We need to add the new clusterRoleBinding to the resourceSet.

The new roleBinding is already covered by the existing resourceSet which matches namespaces:
https://github.com/rancher/backup-restore-operator/blob/f64bef3ed6200dd781860d8dd6b4ccdc2ce1ea26/charts/rancher-backup/files/default-resourceset-contents/rancher.yaml#L25-L27

### DIFF
--- a/charts/rancher-backup/files/default-resourceset-contents/rancher.yaml
+++ b/charts/rancher-backup/files/default-resourceset-contents/rancher.yaml
@@ -27,7 +27,7 @@
   namespaceRegexp: "^cattle-|^p-|^c-|^local$|^user-|^u-"
 - apiVersion: "rbac.authorization.k8s.io/v1"
   kindsRegexp: "^clusterrolebindings$"
-  resourceNameRegexp: "^cattle-|^clusterrolebinding-|^globaladmin-user-|^grb-u-"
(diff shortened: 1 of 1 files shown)


nickgerace (commented):
Don't we also need the `rb-*`?

ryansann:
> Don't we also need the `rb-*` for rolebindings?

I think we're good as we have this block:
```yaml
  kindsRegexp: "^roles$|^rolebindings$"
  namespaceRegexp: "^cattle-|^p-|^c-|^local$|^user-|^u-"
```

where we're using `namespaceRegexp` not `resourceNameRegexp`

nickgerace (approved):

ryansann (approved):

dramich (approved):

Issue #34, opened 2020-09-21
Make sure backup/restore passes CIS hardened profile scan


maggieliu:

jiaqiluo:
This will be covered by [supporting backup/restore installation on hardened cluster ](https://github.com/rancher/backup-restore-operator/issues/29)

jiaqiluo:
can be tested on rancher-backup:`2.0.0+up2.0.0-rc2` in rancher:`master-head`

sowmyav27:
Blocked because of https://github.com/rancher/cis-operator/issues/98

sowmyav27:
**On master-head commit id: `9e0d1b2`**

No additional CIS tests fail on a Hardened local RKE1 cluster after installing rancher backups.

Issue #29, opened 2020-09-16
Support backup/restore installation on hardened cluster


maggieliu:
Support backup/restore installs on hardened cluster

cbron:
This is a requirement for 2.6.

jiaqiluo:
can be tested on rancher-backup:`2.0.0+up2.0.0-rc2` in rancher:`master-head`

sowmyav27:
**On master-head commit id: `d44698d` and backup version `rancher-backup:2.0.0+up2.0.0-rc2`**

- Able to deploy and take a backup on the local cluster - RKE1 which is hardened using [this](https://rancher.com/docs/rancher/v2.x/en/security/rancher-2.5/1.6-hardening-2.5/) doc

PR #99, opened 2021-04-28
add support for installing the app into a hardened cluster

jiaqiluo:
Issues: 
https://github.com/rancher/backup-restore-operator/issues/29
https://github.com/rancher/backup-restore-operator/issues/34

- The chart is tested on a custom hardened cluster under the [CIS 1.6 Benchmark](https://rancher.com/docs/rancher/v2.x/en/security/rancher-2.5/1.6-hardening-2.5/). 
- The app is installed into the hardened cluster successfully.
- After the app is installed, pass the CIS scan with the profile `rke-profile-hardened-1.6`

Next Steps **(after release v2.5.8)** :
- [ ] cut an RC in this repo
- [ ] submit a PR in rancher/chart to cut an RC for the chart

### DIFF
--- a/charts/rancher-backup/templates/hardened.yaml
+++ b/charts/rancher-backup/templates/hardened.yaml
@@ -1,0 +1,116 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "backupRestore.fullname" . }}-patch-sa
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "backupRestore.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-install, post-upgrade
+    "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation
+spec:
+  backoffLimit: 1
+  template:
+    spec:
+      serviceAccountName: {{ include "backupRestore.fullname" . }}-patch-sa
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 1000
+      restartPolicy: Never
+      nodeSelector: {{ include "linux-node-selector" . | nindent 8 }}
+      tolerations: {{ include "linux-node-tolerations" . | nindent 8 }}
+      containers:
+        - name: {{ include "backupRestore.fullname" . }}-patch-sa
+          image: {{ include "system_default_registry" . }}{{ .Values.global.kubectl.repository }}:{{ .Values.global.kubectl.tag }}
+          imagePullPolicy: IfNotPresent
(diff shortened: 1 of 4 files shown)


nickgerace (approved):

charts/rancher-backup/templates/network-policy-allow-all.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: networking.k8s.io/v1
```
aiyengar2, line 1: same comment as https://github.com/rancher/charts/pull/1155#discussion_r623435213. Let's standardize this

aiyengar2 (changes requested):

charts/rancher-backup/templates/network-policy-allow-all.yaml
jiaqiluo: sovled.

nickgerace (approved):
LGTM with Arvind's changes

ryansann (approved):
Looks good

charts/rancher-backup/templates/hardened.yaml
```diff
@@ -0,0 +1,116 @@
...
+      - min: 1
+        max: 65535
+  readOnlyRootFilesystem: false
+  volumes:
+    - 'secret'
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: default-allow-all
```
aiyengar2, line 106: Shouldn't this either contain `{{ include "backupRestore.fullname" . }}` or have a Helm check to not install this NetworkPolicy if it already exists in the `{{ .Release.Namespace }}`?

We would fail if this NetworkPolicy already exists in that namespace right?

aiyengar2 (changes requested):
One minor clarification, otherwise LGTM

charts/rancher-backup/templates/hardened.yaml
jiaqiluo: @aiyengar2 You are right that Helm will fail if the resource exists in the target namespace. I will fix it here.

I see [rancher-monitoring v2](https://github.com/rancher/charts/blob/dac8de7d7e138a5a22da3f45ab9a4168f6d88610/charts/rancher-monitoring/rancher-monitoring/14.5.101%2Bup14.5.0/templates/rancher-monitoring/hardened.yaml#L108-L124) also doesn't check the existences of the NetworkPolicy, do we also need to add checks there?

aiyengar2: yeah we should

PR #101, opened 2021-04-30
bump Golang to 1.16

jiaqiluo:


Dockerfile.dapper
```diff
@@ -1,20 +1,12 @@
-FROM golang:1.13.4-alpine3.10
+FROM golang:1.16.3-alpine3.13
```
cbron, line 1: We have gone both ways, but now I think we should not used tagged images and always build from latest to get security updates

scripts/validate
```diff
@@ -17,5 +15,3 @@ echo Running validation
 echo Running: golangci-lint
 golangci-lint run
 
-echo Running: go fmt
-test -z "$(go fmt ${PACKAGES} | tee /dev/stderr)"
```
cbron, line 21: How come this is removed ?

Dockerfile.dapper
dramich: I think NOT pinning to a version causes some issues: 1. the build will never use a cache since it will always try to pull latest tag from docker. 2. We lose control over what version we are building. If we end up with an issue because we started using a new version of go it will be very hard to figure out where/when that happened. We pin versions to control this and to ensure we are only updating when we know we are doing so.

scripts/validate
dramich: `golangci-lint run` already runs this.

dramich: https://github.com/rancher/backup-restore-operator/blob/master/.golangci.json#L10

Dockerfile.dapper
cbron: Oh whoops, I was thinking this was dockerfile, not dapperfile. We can pin dapperfile.

cbron (approved):
lgtm

PR #110, opened 2021-05-25
Fix: annotations for service account template

Pyunix:
As of current version v1.1.1 using custom annotations for serviceAccount, chart will fail with:
Error: YAML parse error on rancher-backup/templates/serviceaccount.yaml: error converting YAML to JSON: y
aml: line 12: did not find expected key => caused by bad annotations ident(1 space extra, should be inline with labes).


jiaqiluo (approved):
Good catch! Thank you for providing the fix!

Issue #116, opened 2021-05-28
deny all ingress traffic 


jiaqiluo:
Enhancement.

Update the default [`networkPolicy`](https://github.com/rancher/backup-restore-operator/blob/af7d6a1de0ec51b1a51d9dae000f6f097add18d9/charts/rancher-backup/templates/hardened.yaml#L103-L116) to deny all ingress traffic because it does not needed for the app.

jiaqiluo:
can be tested on rancher-backup:`2.0.0+up2.0.0-rc2` in rancher:`master-head`

sowmyav27:
Will be tested as part of this issue - https://github.com/rancher/rancher/issues/33428

PR #119, opened 2021-06-11
block all ingress traffic

jiaqiluo:
issue: https://github.com/rancher/backup-restore-operator/issues/116

reference: https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-deny-all-ingress-traffic

This PR is tested and confirmed working on a hardened cluster.

### DIFF
--- a/charts/rancher-backup/templates/hardened.yaml
+++ b/charts/rancher-backup/templates/hardened.yaml
@@ -107,8 +107,6 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   podSelector: {}
-  ingress:
-    - {}
   egress:
     - {}
   policyTypes:


nickgerace (approved):
Such a small change in YAML to change so much behavior... I'm shocked this works, but the upstream docs do prove it as well. LGTM.

jiaqiluo:
@nickgerace The networkpolicy is a part of the supporting hardening cluster feature that is not released yet, So it will not affect existing users. 
QA will validate that rancher-backup continues working after upgrading. 
We will also document this change.

nickgerace:
@jiaqiluo oh it's not you, im just shocked `- {}` changes so much default behavior.

ryansann (approved):

PR #122, opened 2021-06-18
only replace the image tag for rancher/backup-restore-operator

jiaqiluo:
related issue: https://github.com/rancher/backup-restore-operator/issues/34

We added the image to the value.yaml as part of supporting installation in the hardened cluster, and the image tag of rancher/kube	ctl should not be replaced by the chart version when generating the chart. 

https://github.com/rancher/backup-restore-operator/blob/66a7054f2dfe7ec9e39f4a33f6a485ec266eae1f/charts/rancher-backup/values.yaml#L44-L46

### DIFF
--- a/charts/rancher-backup/values.yaml
+++ b/charts/rancher-backup/values.yaml
@@ -1,6 +1,6 @@
 image:
   repository: rancher/backup-restore-operator
-  tag: v1.0.3
+  tag: %TAG%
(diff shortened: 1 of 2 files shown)


aiyengar2 (approved):

brendarearden (approved):
lgtm

Issue #125, opened 2021-06-21
when creating a backup with S3 and specifying a folder with a trailing `/` instead creates a folder named `/` 


slickwarren:
on master-head (4d1840056b02e390c29a9497abffc840c1de3b28) 
repro steps:
* install rancher backups
* using S3, specify the folder name in your bucket with a trailing `/` i.e. `rancher-backups/`
* create a backup of your cluster

![Screen Shot 2021-06-21 at 9 34 14 AM](https://user-images.githubusercontent.com/16691014/122809117-8c846380-d282-11eb-861c-1a3d42ec173c.png)

Result:
* in S3, a folder in `rancher-backups` is created named `/`. The backup is stored there

![Screen Shot 2021-06-21 at 9 33 10 AM](https://user-images.githubusercontent.com/16691014/122809120-8db59080-d282-11eb-8f2b-24a9ae2ba0f5.png)

jiaqiluo:
can be tested in rancher-backup chart 2.0.0+up2.0.0-rc3

anupama2501:
Verified on master-head b6bcfad

- On an HA server installed rancher backup `rancher-backup:2.0.0+up2.0.0-rc4`
- Following config is provided in the S3 compatible object 

<img width="300" alt="Screen Shot 2021-07-21 at 3 42 24 PM" src="https://user-images.githubusercontent.com/38144301/126569136-748952ca-999a-4aa6-b441-11d78f6c2012.png">
 
- Verified folder `/` is not created and the backup zip file is saved in the folder provided in the config

PR #126, opened 2021-06-25
trim both // inside the path and all leading and trailing /

jiaqiluo:
issue: https://github.com/rancher/backup-restore-operator/issues/125

Note:
This fix only avoid the `//` inside the path that is generated by our code, but it does not handle the case where the user's input has `//`, for example `folder: "a//b//c`

### DIFF
--- a/pkg/controllers/backup/upload.go
+++ b/pkg/controllers/backup/upload.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	v1 "github.com/rancher/backup-restore-operator/pkg/apis/resources.cattle.io/v1"
 	"github.com/rancher/backup-restore-operator/pkg/objectstore"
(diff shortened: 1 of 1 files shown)


pkg/controllers/backup/upload.go
```diff
@@ -24,7 +25,9 @@ func (h *handler) uploadToS3(backup *v1.Backup, objectStore *v1.S3ObjectStore, t
 		if err := os.MkdirAll(filepath.Join(tmpBackupGzipFilepath, objectStore.Folder), os.ModePerm); err != nil {
 			return removeTempUploadDir(tmpBackupGzipFilepath, err)
 		}
-		gzipFile = fmt.Sprintf("%s/%s", objectStore.Folder, gzipFile)
+		// we need to avoid both "//" inside the path and all leading and trailing "/"
+		gzipFile = fmt.Sprintf("%s/%s", strings.TrimRight(objectStore.Folder, "/"), gzipFile)
+		gzipFile = strings.Trim(gzipFile, "/")
```
nickgerace, line 30: Is a TrimRight and a Trim necessary? Would it be better to TrimRight `//` and TrimLeft `/` for readability?

nickgerace: Moreover, small nit, you could probably combine these unless you are worried about line length.

jiaqiluo: the `TrimRight` in the following line is to remove the training `/` in `objectStore.Folder`, such that we will not have  gzipFile = `path//filename`
```
gzipFile = fmt.Sprintf("%s/%s", strings.TrimRight(objectStore.Folder, "/"), gzipFile)
```
the `Trim` in the following line is to remove all leading and trailing `/` and it is equivalent to `TrimLeft` + `TrimRight`
```
gzipFile = strings.Trim(gzipFile, "/")
```


@nickgerace  Please let me know if I answer your question.

nickgerace: Forgive me, I understand they are logically equivalent. I was just curious if you preferred one way over the other for a reason. This looks good to me!

nickgerace (approved):

ryansann (approved):

cbron (approved):

Issue #135, opened 2021-07-20
Custom Resources with empty Status fail restore


snasovich:
Steps to reproduce:
1. Have a backup containing a Custom Resource with empty `status` with Custom Resource Definition actually defining status. For example, at this time `digitaloceanmachinetemplates.rke-machine.cattle.io` can be used as per https://github.com/rancher/rancher/issues/33684
2. Attempt to restore from this backup.

Result:
Restore keeps failing with something like this present in log:
```log
[INFO 2021/07/18 22:11:58] Updating status subresource for "do-to-backup-1-sergey-backup-test-1" of type rke-machine.cattle.io/v1, Resource=digitaloceanmachinetemplates 
[ERRO 2021/07/18 22:11:58] Error restoring resource do-to-backup-1-sergey-backup-test-1 of type rke-machine.cattle.io/v1, Resource=digitaloceanmachinetemplates:
restoreResource: err updating status resource DigitaloceanMachineTemplate.rke-machine.cattle.io "do-to-backup-1-sergey-backup-test-1" is invalid:
status: Invalid value: "null": status in body must be of type object: "null"
```

Expected Behavior:
Custom Resource is restored successfully

Note:
If preparing/massaging backup manually for testing, ensure `filters/statussubresource.json` is included in the backup and lists the resource with the expected status as it's what is used by restore to determine whether attempt is made to restore status.

snasovich:
@jiaqiluo , could you please validate this one? Thank you.

jiaqiluo:
The fix is validated in rancher master-head 2b156fd with the chart `rancher-backup-2.0.0+up2.0.0-rc7`

Note that we need to add the following items to the `resourceSelectors` section of the default `resourceSet` to support the backup of `digitaloceanmachinetemplates` which is used by Fleet. 
```
# Added for v2 provisioning
- apiVersion: apiextensions.k8s.io/v1
  kindsRegexp: .
  resourceNameRegexp: provisioning.cattle.io$|rke-machine-config.cattle.io$|rke-machine.cattle.io$|rke.cattle.io$
- apiVersion: provisioning.cattle.io/v1
  kindsRegexp: .
- apiVersion: rke-machine-config.cattle.io/v1
  kindsRegexp: .
- apiVersion: rke-machine.cattle.io/v1
  kindsRegexp: .
- apiVersion: rke.cattle.io/v1
  kindsRegexp: .
- apiVersion: apiextensions.k8s.io/v1
[...]

PR #136, opened 2021-07-22
Fix for restoring resources with empty status

snasovich:
If CRD defines `status` but CR resource to be restored doesn't have one, the restore fails. This change ensures that only non-empty status is attempted to be restored.

For #135

### DIFF
--- a/pkg/controllers/restore/controller.go
+++ b/pkg/controllers/restore/controller.go
@@ -574,7 +574,7 @@ func (h *handler) restoreResource(restoreObjInfo objInfo, restoreObjData unstruc
 		if err != nil {
 			return err
 		}
-		if hasStatusSubresource {
+		if hasStatusSubresource && obj.Object["status"] != nil {
 			logrus.Infof("Post-create: Updating status subresource for %#v of type %v", name, gvr)
 			createdObj.Object["status"] = obj.Object["status"]
 			_, err := dr.UpdateStatus(h.ctx, createdObj, k8sv1.UpdateOptions{})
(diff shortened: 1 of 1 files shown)


pkg/controllers/restore/controller.go
```diff
@@ -574,7 +574,7 @@ func (h *handler) restoreResource(restoreObjInfo objInfo, restoreObjData unstruc
 		if err != nil {
 			return err
 		}
-		if hasStatusSubresource {
+		if hasStatusSubresource && obj.Object["status"] != nil {
```
nickgerace, line 577: What if we want to restore CRs without a status? Does this still work?

snasovich: I can test, but I imagine `hasStatusSubresource` will be `false` in this case and then block wasn't running before the fix and won't be running following the fix.

nickgerace: If that's the case, then this LGTM. We should confirm first how/when that's seet.

snasovich: @nickgerace , just confirmed that all kinds of resources (ones that don't have status, ones that do have status and define it, and ones that do have status but do not define it) import correctly.
I'm not sure I understand what you mean by _"we can add your new logic to the function that sets it so that it's co-located"_ - there is really not much of new logic here. We're simply ensuring restore doesn't attempt to set status if it's not available on the resource. Normally this should not happen but we do have some resources now (due to https://github.com/rancher/rancher/issues/33684) that have empty status.

nickgerace: Sounds good, I will resolve this.

Ah I can explain. I am suggesting that we may want to see if we can move the `obj.Object["status"] != nil` check to where the boolean is set. If the `status` field is nil, then we set `hasStatusSubresource` to `false.`
https://github.com/rancher/backup-restore-operator/blob/c35f3bc1ecc749cf98344f7e699bb778ead4a6d6/pkg/controllers/restore/controller.go#L505

snasovich: IMO either way is fine, but I personally lean towards how it's implemented. This flag is more like "does CRD define the status?" and it's responsibility of `restoreResource` func to restore status if it's available. But I see how there could be valid arguments for what you're suggesting. Given that you're fine with the original approach, I would prefer to keep it.

nickgerace: Alright let's do that then! Fair points.

nickgerace (approved):

jiaqiluo (approved):
LGTM

PR #137, opened 2021-07-28
remove the trailing slash from the folder name

jiaqiluo:
https://github.com/rancher/rancher/issues/33734

Problem:
If the value of the `folder name` contains trailing slash `/`, the full path to the backup file in S3 will have double slashes in it, which leads to the `no backups found` error 

Fix:
Trim the trailing slash from the value. 

Next Step:

- [ ] cut a new tag 
- [ ] update the rancher-backup chart


nickgerace (approved):
LGTM, but we should be considering using `filepath.Join` to handle path seps

snasovich (approved):

cbron (approved):

PR #138, opened 2021-07-29
fix missing token secret error 

jiaqiluo:
https://github.com/rancher/rancher/issues/33794

Issue:
The token secret will be deleted if the service account it binds to does not exist. It will always fail to find the service account because the UID of the service account is changed (the cluster assigns a new UID to it).

But the SA does not know about it, so once the SA is restored, it will link 2 secrets, one's name is the one in the backup, another is the new one created by Token Controller. The 1st one is missing, so the following error happens
```
[FATAL] secrets "rancher-token-48bj8" not found
``` 

~this issue seems related to the new feature [BoundServiceAccountTokenVolume](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#bound-service-account-token-volume) which is enabled by default in 1.21~ 
The issue is found in a 1.21 cluster but it looks like a long-time existing bug that we just did not hit in the past. 

Solution:
ignore the secrets section in the service account to avoid missing token secret errors in k8s 1.21
so we just let the cluster auto-create a new token secret for the SA. 

Note:
I have tested on two rke 1.21 clusters that rancher (using `master-head` image) is backed up and restored successfully to a new cluster. 

TODO after the PR is merged:

- [ ] cut a new tag
- [ ] update chart in rancher/charts

### DIFF
--- a/pkg/controllers/restore/controller.go
+++ b/pkg/controllers/restore/controller.go
@@ -41,6 +41,7 @@ const (
 	clusterScoped   = "clusterscoped"
 	namespaceScoped = "namespaceScoped"
 	leaseName       = "restore-controller"
+	secretsMapKey   = "secrets"
 )
 
 type handler struct {
@@ -369,6 +370,11 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
 		}
 
 		metadata := resourceData.Object[metadataMapKey].(map[string]interface{})
+		// remove secrets section as referenced secrets will be removed by k8s Token Controller as they are considered orphaned
+		if kind, ok := resourceData.Object["kind"]; ok && kind == "ServiceAccount" {
+			delete(currRestoreObj.Data.Object, secretsMapKey)
+			logrus.Debugf("Secrets section is ignored in ServiceAccount %s", name)
+		}
 		ownerRefs, ownerRefsFound := metadata[ownerRefsMapKey].([]interface{})
 		if !ownerRefsFound {
 			// has no owners, so no need to add to adjacency list, add to restoreResources list


pkg/controllers/restore/controller.go
```diff
@@ -369,6 +370,10 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
 		}
 
 		metadata := resourceData.Object[metadataMapKey].(map[string]interface{})
+		// ignore the secrets section in the service account to avoid missing token secret error in k8s 1.21
```
snasovich, line 373: I don't think it's k8s 1.21-specific. I've had this issue on local k3d cluster (`v1.19.8+k3s1`). To test, output any SA to yaml and then apply this yaml to another cluster. I bet it will have 2 secrets linked by it.

pkg/controllers/restore/controller.go
```diff
@@ -369,6 +370,10 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
...
+		if resourceData.Object["kind"].(string) == "ServiceAccount" {
+			delete(currRestoreObj.Data.Object, secretsMapKey)
```
snasovich, line 375: Looking at func name/description `generateDependencyGraph` I would not expect it to "massage" objects. If there is already an existing function that does some sort of sanitize of objects to restore, I would add it there. If not, maybe worth introducing a separate func for this?

snasovich: Maybe worth logging this happened? At least at Debug level, if not Info.

snasovich (changes requested):

pkg/controllers/restore/controller.go
jiaqiluo: I put it here because the code below (within the same method) also modifies the object: correcting the `ownerRefs`. 

We should move both of them to a new method, but I'd like to leave it this way for now. 

A debug messaging is added right below this line.

pkg/controllers/restore/controller.go
jiaqiluo: the inline comment is updated.

pkg/controllers/restore/controller.go
snasovich: @jiaqiluo , thought about this as well, but modifying owner references is much more warranted in a function named `generateDependencyGraph` - one could expect owner references need massaging here; not so much with removing secret.
I guess you could argue secret is a dependency of an SA. I'm OK with not addressing it overall.

pkg/controllers/restore/controller.go
snasovich: Ideally we want description of the PR updated too. :)

jiaqiluo: done.

snasovich (approved):

pkg/controllers/restore/controller.go
```diff
@@ -369,6 +370,11 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
 		}
 
 		metadata := resourceData.Object[metadataMapKey].(map[string]interface{})
+		// ignore the secrets section in the service account to avoid missing token secret error
+		if resourceData.Object["kind"].(string) == "ServiceAccount" {
```
cbron, line 374: So this will actually get hit on every type of object, but then if the object type is a service account then we delete the secret that backs it ?

pkg/controllers/restore/controller.go
```diff
@@ -369,6 +370,11 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
...
+			delete(currRestoreObj.Data.Object, secretsMapKey)
+			logrus.Debugf("The secret section is ignored from the service account %s", name)
```
cbron, line 376:
```suggestion
			logrus.Debugf("ServiceAccount secret ignored  %s", name)
```

pkg/controllers/restore/controller.go
jiaqiluo: correct.

pkg/controllers/restore/controller.go
snasovich: This reads to be that secret name follows rather than SA's name. `ServiceAccount secret ignored, SA name: %s` maybe?

pkg/controllers/restore/controller.go
```diff
@@ -369,6 +370,11 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
 		}
 
 		metadata := resourceData.Object[metadataMapKey].(map[string]interface{})
+		// ignore the secrets section in the service account to avoid missing token secret error
```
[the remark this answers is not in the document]
snasovich: `// remove secrets section as referenced secrets will be removed by k8s Token Controller as they are considered orphaned` maybe?

pkg/controllers/restore/controller.go
snasovich: As discussed in chat, type assertion is unnecessary here. In fact it may result in panic is somehow it's not string or nil.

pkg/controllers/restore/controller.go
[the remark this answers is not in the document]
jiaqiluo: done.

pkg/controllers/restore/controller.go
jiaqiluo: done.

nickgerace (approved):
This was a gnarly one. Nice work

ryansann (approved):
LGTM

snasovich (approved):

cbron (approved):

PR #141, opened 2021-08-04
Fix for setting backup and restore status

snasovich:
https://github.com/rancher/rancher/issues/33946
and
https://github.com/rancher/rancher/issues/33942

For restore object, just needed to add `omitempty`. Tested by running local against `v1.20.9+rke2r1` setup provided by Jan Bruder who reported the issue.
Note missing `omitempty` on `ObservedGeneration` doesn't cause any failures, but having it set to 0 is rather unsightly so I've updated it too.

As far as backup object goes, if there is a failure none of direct Status fields are set, so all of them got `omitempty` added.

TODO after the PR is merged:
- [ ] Cut a new tag
- [ ] Update chart in rancher/charts

### DIFF
--- a/pkg/apis/resources.cattle.io/v1/types.go
+++ b/pkg/apis/resources.cattle.io/v1/types.go
@@ -37,13 +37,13 @@ type BackupSpec struct {
 
 type BackupStatus struct {
 	Conditions         []genericcondition.GenericCondition `json:"conditions"`
-	LastSnapshotTS     string                              `json:"lastSnapshotTs"`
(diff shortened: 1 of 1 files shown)


nickgerace (approved):

dramich (approved):

nickgerace (approved):

jiaqiluo (approved):
LGTM

snasovich:
This is actually an issue with this particular version of RKE2, see https://github.com/rancher/rancher/issues/33946#issuecomment-893832907 for explanation. Closing this PR without merging.

PR #144, opened 2021-08-07
Fix backup for fleet

StrongMonkey:
Fleet cluster needs to be reimported in order to reissue service account token that is no longer valid during backup and restore.

https://github.com/rancher/rancher/issues/33954


charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
```diff
@@ -4,6 +4,7 @@
 - apiVersion: "v1"
   kindsRegexp: "^secrets$"
   namespaceRegexp: "^cattle-fleet-|^fleet-|^cluster-fleet-"
+  resourceNameRegexp:
```
snasovich, line 7: Why are we adding empty regexp here?

pkg/controllers/restore/controller.go
```diff
@@ -477,6 +476,26 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
...
+func customize(obj *unstructured.Unstructured) {
+	switch obj.GetKind() {
+	case "ServiceAccount":
+		// remove secrets section as referenced secrets will be removed by k8s Token Controller as they are considered orphaned
+		delete(obj.Object, secretsMapKey)
+		logrus.Debugf("Secrets section is ignored in ServiceAccount %s/%s", obj.GetNamespace(), obj.GetName())
+	case "Cluster":
+		// for fleet cluster it needs to be reimported in order to reissue service account token that is no longer valid
+		if obj.GetAPIVersion() == "fleet.cattle.io/v1alpha1" {
+			n, err := rand.Int(rand.Reader, big.NewInt(99999))
```
snasovich, line 489: I know it's one of those "never going to happen", but technically `err` may be non-nil and we simply ignore it here with no logging or actual error.

pkg/controllers/restore/controller.go
```diff
@@ -477,6 +476,26 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
...
+			if err == nil {
+				if err := unstructured.SetNestedField(obj.Object, n.Int64(), "spec", "redeployAgentGeneration"); err != nil {
```
snasovich, line 491: What exactly is `redeployAgentGeneration` and why is it acceptable to set random value to it? Assuming it's already set to some value, what if random happens to return the same value as well?

pkg/controllers/restore/controller.go
```diff
@@ -477,6 +476,26 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
...
+					logrus.Warnf("Fleet cluster %s/%s can't be reset to be re-imported", obj.GetNamespace(), obj.GetName())
```
snasovich, line 492: I would argue this should be considered a true error and fail import of this object (and consequently the import itself as well). If we somehow restore without this change to CR applied, restored cluster won't be functioning OK.

snasovich (changes requested):

pkg/controllers/restore/controller.go
StrongMonkey: It will redeploy agent in downstream cluster which will reissue serviceaccount token. Yeah will be good if we check if value matches and regenerate it.

pkg/controllers/restore/controller.go
StrongMonkey: Yeah, I actually don't want this error to fail the whole restore process so that in case error happens, user can patch it manually. This will not likely to cause an error unless the fields are incorrect so I think printing warning might be enough.

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
StrongMonkey: Typo, will remove :+1:

pkg/controllers/restore/controller.go
StrongMonkey: I am going to take +1 approach here instead of random int, seems overkill.

pkg/controllers/restore/controller.go
StrongMonkey: It is more like conservative so that in case we break the api that this is causing error, user has way to workaround it instead of letting it fail.

snasovich: I see the logic here, but a bit concerned users won't know which message to check for. Though given how unlikely is this error to happen, I think we're OK with the current approach.

pkg/controllers/restore/controller.go
```diff
@@ -477,6 +474,31 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
...
+		// remove secrets section as referenced secrets will be removed by k8s Token Controller as they are considered orphaned
+		delete(obj.Object, secretsMapKey)
+		logrus.Debugf("Secrets section is ignored in ServiceAccount %s/%s", obj.GetNamespace(), obj.GetName())
+	case "Cluster":
+		// for fleet cluster it needs to be reimported in order to reissue service account token that is no longer valid
+		if obj.GetAPIVersion() == "fleet.cattle.io/v1alpha1" {
+			// only warn error if field can't be found. In case there are breaking api changes error will be printed as Warn and user has workaround to patch it.
+			redeployAgentGeneration, ok, err := unstructured.NestedInt64(obj.Object, "spec", "redeployAgentGeneration")
+			if err != nil {
+				logrus.Warnf("Fleet cluster %s/%s can't be reset to be re-imported, failed to fetch .spec.redeployAgentGeneration", obj.GetNamespace(), obj.GetName())
```
snasovich, line 490: Recommend adding `err` message to the message we're logging as otherwise we don't know why it happened. Same for the other WARN.

pkg/controllers/restore/controller.go
```diff
@@ -477,6 +474,31 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
...
+				return
+			}
+			if ok {
```
snasovich, line 493: I think we need to handle `ok == false` case as well. On the object that I have on my environment this field is not set:

```
> kc get clusters.fleet.cattle.io -n fleet-default do-rke2-backup-test -o yaml                                                                        

apiVersion: fleet.cattle.io/v1alpha1
kind: Cluster
metadata:
  annotations:
...
spec:
  clientID: wzpsf4272bqbwln7tgcqpt2ttmftwj7fz9ps9zc95v9kdqgmkmsmnj
  kubeConfigSecret: do-rke2-backup-test-kubeconfig
status:
  agent:
...
```

snasovich (changes requested):

pkg/controllers/restore/controller.go
StrongMonkey: Good point. There is no need to check ok.

snasovich (approved):
LGTM

jiaqiluo (approved):
LGTM

PR #147, opened 2021-08-19
remove the deployment in ns cattle-system from the resourceSet

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/34247

Fixes:     
- remove deployment in  ns `cattle-system` from the ResourceSet, so a new backup does not include rancher and rancher-webhook
- skip the deployment rancher and rancher-webhook when restoring

With the fixies, we can get the following results:

 

 If deployment rancher is in the backup | migrate to a new cluster (no rancher exists)  | restore in the same cluster (rancher exists) | rollback in the same cluster ( upgraded rancher exists)
-- | -- | -- | --
yes | no rancher | existing rancher runs up | existing rancher 2.6.0 runs up *can be fixed by new steps
no  | no rancher | existing rancher runs up | existing rancher 2.6.0 runs up *can be fixed by new steps
 
The case of `rollback in the same cluster ( upgraded rancher exists)` is based on this scenario:
```
- install Rancher v2.5.9 into the cluster
- make a backup
- upgrade rancher to 2.6.0
- rollback to 2.5.9 by restoring the backup
```

The above 2 cases with * can be solved by following the new steps for rolling back rancher
```
- scale the deployment rancher to 0 
- - restore the backup file
- run `helm rollback rancher -n cattle-system` to run rancher 2.5.9 up 
```
A separate issue for updating the documentation is made: https://github.com/rancher/docs/issues/3460

TODO after mering:
- [] back port to the 2.5 line

### DIFF
--- a/pkg/controllers/restore/controller.go
+++ b/pkg/controllers/restore/controller.go
@@ -356,10 +356,15 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
 		resourceInfoToData = objFromBackupCR.namespacedResourceInfoToData
 	}
 	for resourceInfo, resourceData := range resourceInfoToData {
-		// add to adjacency list
 		name := resourceInfo.Name
 		namespace := resourceInfo.Namespace
 		gvr := resourceInfo.GVR
+		if resourceData.GetKind() == "Deployment" && namespace == "cattle-system" {
+			if strings.HasSuffix(name, "rancher") || strings.HasSuffix(name, "rancher-webhook") {
+				logrus.Infof("Skip restoring the deployment %s/%s", namespace, name)
+				continue
+			}
+		}
 		// TODO: Maybe restoreObj won't be needed
 		currRestoreObj := restoreObj{
 			Name:               name,
--- a/charts/rancher-backup/files/default-resourceset-contents/rancher.yaml
+++ b/charts/rancher-backup/files/default-resourceset-contents/rancher.yaml
@@ -3,10 +3,6 @@
   resourceNameRegexp: "^cattle-|^p-|^c-|^user-|^u-"
   resourceNames:
     - "local"
-- apiVersion: "apps/v1"
-  kindsRegexp: "^deployments$"
-  namespaces:
-    - "cattle-system"
 - apiVersion: "v1"
   kindsRegexp: "^secrets$"
   namespaceRegexp: "^cattle-|^p-|^c-|^local$|^user-|^u-"


pkg/controllers/restore/controller.go
```diff
@@ -370,6 +370,12 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
 		}
 
 		customize(&resourceData)
+		if resourceData.GetKind() == "Deployment" && namespace == "cattle-system" {
```
snasovich, line 373: This check can be moved all the way as the first thing we do in this loop.

pkg/controllers/restore/controller.go
```diff
@@ -370,6 +370,12 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
...
+			if strings.HasSuffix(name, "rancher") || strings.HasSuffix(name, "rancher-webhook") {
```
snasovich, line 374: Are there cases where Rancher / webhook deployments are named anything different than `rancher` and `rancher-webhook` exactly? If not, we don't need `HasSuffix` checks.

snasovich (changes requested):

pkg/controllers/restore/controller.go
jiaqiluo: Currently, we have only those 2. 
I added the check to avoid deleting others if we have them in the future. 

Should I  not worry about it now?

pkg/controllers/restore/controller.go
jiaqiluo: moved.

pkg/controllers/restore/controller.go
snasovich: `name == "rancher"` is more restrictive than `strings.HasSuffix(name, "rancher")` - "rancher-something" is false for the former and true for the latter.
So, if we want to avoid deleting others - we should go with more restrictive check.

I was wondering if you went with `HasSuffix` because you know of deployments named like `rancher-something` that should also be skipped. Looks like it's not the case.

pkg/controllers/restore/controller.go
```diff
@@ -360,6 +360,12 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
 		name := resourceInfo.Name
```
snasovich, line 360: `// add to adjacency list` comment was already somewhat out of place and it's even more so now after we added skip logic so close to it. Suggesting to remove this comment as it's largely redundant.

pkg/controllers/restore/controller.go
jiaqiluo: If the user sets a different name when installing rancher, for example, `helm install my-setup rancher-latest/rancher`, the deployment’s name will be `my-setup-rancher`. `strings.HasSuffix(name, "rancher")` is to catch this case.

pkg/controllers/restore/controller.go
jiaqiluo: removed.

pkg/controllers/restore/controller.go
snasovich: Thank you for the explanation, makes sense. Also, my example in comment was incorrect (suffix vs. prefix mix-up), I've corrected it.

snasovich (approved):
LGTM

pkg/controllers/restore/controller.go
```diff
@@ -356,10 +356,15 @@ func (h *handler) generateDependencyGraph(ownerToDependentsList map[string][]res
 		resourceInfoToData = objFromBackupCR.namespacedResourceInfoToData
 	}
 	for resourceInfo, resourceData := range resourceInfoToData {
-		// add to adjacency list
 		name := resourceInfo.Name
 		namespace := resourceInfo.Namespace
 		gvr := resourceInfo.GVR
+		if resourceData.GetKind() == "Deployment" && namespace == "cattle-system" {
+			if strings.HasSuffix(name, "rancher") || strings.HasSuffix(name, "rancher-webhook") {
```
nickgerace, line 363: Why use a substring comparison over a direct one?

jiaqiluo: If the user sets a different name when installing rancher, for example, `helm install my-setup rancher-latest/rancher`, the deployment’s name will be `my-setup-rancher`. `strings.HasSuffix(name, "rancher")` is to catch this case. 
So does `strings.HasSuffix(name, "rancher-webhook")`

nickgerace (approved):

cbron (approved):
The code here doesn't have a ton to do with the chart, this code really just stops rancher deployment from getting restored in either case. The docs issue is the real fix to the chart permutations. But that issues is made now, so looks good.

PR #148, opened 2021-08-21
Do not store kubeconfig token, add machine secret

StrongMonkey:
https://github.com/rancher/rancher/issues/33954

Context: https://github.com/rancher/rancher/issues/33954#issuecomment-903001616

TLDR: we shouldn't be restoring token associated with kubeconfig secret, as it is cached by secret and we can't restore secret too because it has rancher service IP which could change during restore. Also add machine secret so that during a restore ssh and download keys are still available for machine nodes.

### DIFF
--- a/charts/rancher-backup/files/default-resourceset-contents/provisioningv2.yaml
+++ b/charts/rancher-backup/files/default-resourceset-contents/provisioningv2.yaml
@@ -13,6 +13,6 @@
   kindsRegexp: "."
 - apiVersion: "v1"
   kindsRegexp: "^secrets$"
-  resourceNameRegexp: "machine-plan$|rke-state$"
+  resourceNameRegexp: "machine-plan$|rke-state$|machine-state$|machine-driver-secret$|machine-provision$"
   namespaces:
   - "fleet-default"
--- a/charts/rancher-backup/files/default-resourceset-contents/rancher.yaml
+++ b/charts/rancher-backup/files/default-resourceset-contents/rancher.yaml
@@ -36,6 +36,15 @@
   resourceNameRegexp: "management.cattle.io$|project.cattle.io$|catalog.cattle.io$|resources.cattle.io$"
 - apiVersion: "management.cattle.io/v3"
   kindsRegexp: "."
+  excludeKinds:
+    - "tokens"
+- apiVersion: "management.cattle.io/v3"
+  kindsRegexp: "^tokens$"
+  labelSelectors:
+    matchExpressions:
+      - key: "authn.management.cattle.io/kind"
+        operator: "NotIn"
+        values: [ "provisioning" ]
 - apiVersion: "project.cattle.io/v3"
   kindsRegexp: "."
 - apiVersion: "catalog.cattle.io/v1"
--- a/pkg/resourcesets/collector.go
+++ b/pkg/resourcesets/collector.go
@@ -134,11 +134,20 @@ func (h *ResourceHandler) gatherResourcesForGroupVersion(filter v1.ResourceSelec
 	// "resources" list has all resources under given groupVersion, first filter based on KindsRegexp
 	if filter.KindsRegexp != "" {
 		if filter.KindsRegexp == "." {
-			// "." will match anything, so return entire resource list
-			return resources.APIResources, nil
+			// "." will match anything except the ones in the list of excludeKinds
+			for _, res := range resources.APIResources {
+				if isKindExcluded(filter.ExcludeKinds, res) {
+					continue
+				}
+				resourceListFromRegex = append(resourceListFromRegex, res)
+			}
+			return resourceListFromRegex, nil
 		}
 		// else filter out resource with regex match
 		for _, res := range resources.APIResources {
+			if isKindExcluded(filter.ExcludeKinds, res) {
+				continue
+			}
 			kindMatched, err := regexp.MatchString(filter.KindsRegexp, res.Kind)
 			if err != nil {
 				return resourceList, err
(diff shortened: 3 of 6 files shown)


charts/rancher-backup/files/default-resourceset-contents/rancher.yaml
```diff
@@ -36,6 +36,15 @@
   resourceNameRegexp: "management.cattle.io$|project.cattle.io$|catalog.cattle.io$|resources.cattle.io$"
 - apiVersion: "management.cattle.io/v3"
   kindsRegexp: "."
+  excludeKinds:
```
StrongMonkey, line 39: For some reason go regxp doesn't support exclude, and also it might be quite expensive to support that. So I am just adding this.

pkg/resourcesets/collector.go
```diff
@@ -134,11 +134,20 @@ func (h *ResourceHandler) gatherResourcesForGroupVersion(filter v1.ResourceSelec
 	// "resources" list has all resources under given groupVersion, first filter based on KindsRegexp
 	if filter.KindsRegexp != "" {
 		if filter.KindsRegexp == "." {
-			// "." will match anything, so return entire resource list
-			return resources.APIResources, nil
+			// "." will match anything expect excludeKinds, so return entire resource list
+			for _, res := range resources.APIResources {
+				if slice.ContainsString(filter.ExcludeKinds, res.Kind) {
```
jiaqiluo, line 139: Although the value of Kind is in the PascalCasing format (capitalizing the first letter of each word) by design, should we consider some variations in users' input? For example, "Token" vs "token", "ConfigMap" vs "configMap" vs "configmap". 

A possible solution for it is to capitalize values of `filter.ExcludeKinds` and `res.Kind` before comparing. 

FYI, we support both PascalCasing and all lower letters in the `kindsRegexp`.

jiaqiluo (changes requested):

charts/rancher-backup/files/default-resourceset-contents/rancher.yaml
snasovich: Yup, Go's flavor of regexp doesn't support negative lookahead which makes is (nearly?) impossible to build regexps that match "anything except <word1>, <word2>, ...": https://stackoverflow.com/a/26792316/16564280
Just wanted ti share the link for more context.

pkg/resourcesets/collector.go
snasovich: I think we're case-sensitive in all other cases, so it would be weird that we're case insensitive here. I do see that we match `KindsRegexp` to both `res.Kind` and `res.Name`, so should we check both of these against `filter.ExcludeKinds` if the latter is specified?

Another point towards this is how weird this looks in config:
```yaml
...
excludeKinds:
    - "Token"
- apiVersion: "management.cattle.io/v3"
  kindsRegexp: "^tokens$"
...
```
So we have `excludeKinds` set as `Token` and right below it we have `kindsRegexp` as `^tokens$` making it look quite inconsistent.

StrongMonkey: I can make it case-insentive. But I agree with @snasovich that it does looks weird on the config, but there is no good way to make an exception match for a type today

snasovich: @StrongMonkey , I was suggesting to have `excludeKinds` act on `res.Name` as well to match behavior of `kindsRegexp`. Still a bit questionable behavior, but it would make it more consistent and allow us to configure.

On a side note we could do `excludeKindsRegexp` string to have more flexibility compared to current `excludeKinds` array. But I'll leave it to you to decide if we want to change to it.

StrongMonkey: @snasovich @jiaqiluo ok, I added support for both kind and plural names so it looks more consistent now. And the config is also changed to use plural name of token. I will skip the implementation of `excludeKindsRegexp` for now since it is just a hot fix and I would like to avoid any feature work that increase more complexity, but I do agree that is a useful feature if we are going to add it later.

pkg/resourcesets/collector.go
```diff
@@ -134,11 +134,20 @@ func (h *ResourceHandler) gatherResourcesForGroupVersion(filter v1.ResourceSelec
 	// "resources" list has all resources under given groupVersion, first filter based on KindsRegexp
 	if filter.KindsRegexp != "" {
 		if filter.KindsRegexp == "." {
-			// "." will match anything, so return entire resource list
-			return resources.APIResources, nil
+			// "." will match anything expect excludeKinds, so return entire resource list
```
jiaqiluo, line 137:
```suggestion
			// "." will match anything except the ones in the list of excludeKinds

**Review by jiaqiluo — CHANGES_REQUESTED** [review]

**Review by snasovich — APPROVED** [review]
LGTM

**Review by jiaqiluo — APPROVED** [review]
LGTM

@StrongMonkey Will the associated issue happen in rancher 2.5.x? If so, we need to backport this fix to the `release/v1.0` branch in this repo.

**Comment by StrongMonkey** [comment]
@jiaqiluo No this is specific to 2.6. Related to v2 provisioning and rke2

File: charts/rancher-backup/files/default-resourceset-contents/provisioningv2.yaml [hunk 4]
```diff
@@ -13,6 +13,6 @@
   kindsRegexp: "."
 - apiVersion: "v1"
   kindsRegexp: "^secrets$"
-  resourceNameRegexp: "machine-plan$|rke-state$"
+  resourceNameRegexp: "machine-plan$|rke-state$|machine-state$|machine-driver-secret$|machine-provision$"
```
cbron, line 16: so this is unrelated

charts/rancher-backup/files/default-resourceset-contents/rancher.yaml
```diff
@@ -36,6 +36,15 @@
...
+    - "tokens"
```
cbron, line 40: Won't this invalidate all kubeconfigs on the cluster ? And anything else using tokens ? What does the section below do ?

cbron: Talked to Daishan: this part backs up everything _but_ tokens, the part below it backs up all tokens except those with provisioning label. So user kubeconfigs are fine.

cbron (approved):

PR #149, opened 2021-08-22
New way to handle subResource Status 

jiaqiluo:
issue: https://github.com/rancher/rancher/issues/34193

drop the file `statussubresource.json` and implement a new approach to restore resources the subresource Status

Bug:
While going through the collected resources, the current implementation checks if its name is `xxx/status` and supports the update operation; if so, add its parent resource kind to a list that is saved to a file. And the list is used as the source of truth for determining if the status of a resource needs to be updated. 
And its problem is that if the rules in the resourceSet are strict, we will miss the status resource, therefore not update the status after restoring. 

Fix:
The new approach: when restoring the CRDs, if we see the `spec.subresources.status` filed, it means this kind of resource has the status subresource that needs to be restored. We save the names of those CRD into a list, then when restoring other resources, we will update its status if its kind is in the list. 
The new approach is more robust as it checks the CRD directly instead of depending on collected resources. 

TODO:

- [ ] backport to  the 2.5 line

### DIFF
--- a/pkg/controllers/restore/controller.go
+++ b/pkg/controllers/restore/controller.go
@@ -16,6 +16,7 @@ import (
 	v1core "github.com/rancher/wrangler-api/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/condition"
 	"github.com/rancher/wrangler/pkg/genericcondition"
+	"github.com/rancher/wrangler/pkg/slice"
 	"github.com/sirupsen/logrus"
 
 	coordinationv1 "k8s.io/api/coordination/v1"
@@ -36,12 +37,14 @@ import (
 )
 
 const (
-	metadataMapKey  = "metadata"
-	ownerRefsMapKey = "ownerReferences"
-	clusterScoped   = "clusterscoped"
-	namespaceScoped = "namespaceScoped"
-	leaseName       = "restore-controller"
-	secretsMapKey   = "secrets"
+	metadataMapKey     = "metadata"
+	ownerRefsMapKey    = "ownerReferences"
+	clusterScoped      = "clusterscoped"
+	namespaceScoped    = "namespaceScoped"
+	leaseName          = "restore-controller"
+	secretsMapKey      = "secrets"
+	specMapKey         = "spec"
+	subResourcesMapKey = "subresources"
 )
 
 type handler struct {
@@ -64,7 +67,6 @@ type ObjectsFromBackupCR struct {
 	clusterscopedResourceInfoToData map[objInfo]unstructured.Unstructured
 	namespacedResourceInfoToData    map[objInfo]unstructured.Unstructured
 	resourcesFromBackup             map[string]bool
-	resourcesWithStatusSubresource  map[string]bool
 	backupResourceSet               v1.ResourceSet
 }
 
@@ -140,6 +142,7 @@ func (h *handler) OnRestoreChange(_ string, restore *v1.Restore) (*v1.Restore, e
 
 	created := make(map[string]bool)
 	ownerToDependentsList := make(map[string][]restoreObj)
+	var crdsWithSubStatus []string
 	var toRestore []restoreObj
 	numOwnerReferences := make(map[string]int)
 	objFromBackupCR := ObjectsFromBackupCR{
@@ -147,7 +150,6 @@ func (h *handler) OnRestoreChange(_ string, restore *v1.Restore) (*v1.Restore, e
 		clusterscopedResourceInfoToData: make(map[objInfo]unstructured.Unstructured),
 		namespacedResourceInfoToData:    make(map[objInfo]unstructured.Unstructured),
 		resourcesFromBackup:             make(map[string]bool),
-		resourcesWithStatusSubresource:  make(map[string]bool),
 		backupResourceSet:               v1.ResourceSet{},
 	}
 
@@ -213,7 +215,7 @@ func (h *handler) OnRestoreChange(_ string, restore *v1.Restore) (*v1.Restore, e
 
 	// first restore CRDs
 	logrus.Infof("Starting to restore CRDs for restore CR %v", restore.Name)
-	if err := h.restoreCRDs(created, objFromBackupCR); err != nil {
+	if crdsWithSubStatus, err = h.restoreCRDs(created, objFromBackupCR); err != nil {
 		h.scaleUpControllersFromResourceSet(objFromBackupCR)
 		logrus.Errorf("Error restoring CRDs %v", err)
 		// Cannot set the exact error on reconcile condition, the order in which resources failed to restore are added in err msg could
@@ -223,7 +225,7 @@ func (h *handler) OnRestoreChange(_ string, restore *v1.Restore) (*v1.Restore, e
 
 	logrus.Infof("Starting to restore clusterscoped resources for restore CR %v", restore.Name)
 	// then restore clusterscoped resources, by first generating dependency graph for cluster scoped resources, and create from the graph
-	if err := h.restoreClusterScopedResources(ownerToDependentsList, &toRestore, numOwnerReferences, created, objFromBackupCR); err != nil {
+	if err := h.restoreClusterScopedResources(ownerToDependentsList, &toRestore, numOwnerReferences, created, objFromBackupCR, crdsWithSubStatus); err != nil {
 		h.scaleUpControllersFromResourceSet(objFromBackupCR)
 		logrus.Errorf("Error restoring cluster-scoped resources %v", err)
 		return h.setReconcilingCondition(restore, fmt.Errorf("error restoring cluster-scoped resources, check logs for exact error"))
@@ -233,7 +235,7 @@ func (h *handler) OnRestoreChange(_ string, restore *v1.Restore) (*v1.Restore, e
 	// now restore namespaced resources: generate adjacency lists for dependents and ownerRefs for namespaced resources
 	ownerToDependentsList = make(map[string][]restoreObj)
 	toRestore = []restoreObj{}
-	if err := h.restoreNamespacedResources(ownerToDependentsList, &toRestore, numOwnerReferences, created, objFromBackupCR); err != nil {
+	if err := h.restoreNamespacedResources(ownerToDependentsList, &toRestore, numOwnerReferences, created, objFromBackupCR, crdsWithSubStatus); err != nil {
 		h.scaleUpControllersFromResourceSet(objFromBackupCR)
 		logrus.Errorf("Error restoring namespaced resources %v", err)
 		return h.setReconcilingCondition(restore, fmt.Errorf("error restoring namespaced resources, check logs for exact error"))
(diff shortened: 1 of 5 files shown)


pkg/controllers/restore/controller.go
```diff
@@ -273,13 +275,25 @@ func (h *handler) OnRestoreChange(_ string, restore *v1.Restore) (*v1.Restore, e
...
 	for crdInfo, crdData := range objFromBackupCR.crdInfoToData {
 		err := h.restoreResource(crdInfo, crdData, false)
 		if err != nil {
 			return fmt.Errorf("restoreCRDs: %v", err)
 		}
 		created[crdInfo.ConfigPath] = true
+
+		specs := crdData.Object[specMapKey].(map[string]interface{})
+		subResources, ok := specs[subResourcesMapKey]
+		if ok {
```
nickgerace, line 288: If this is not ok or nil, we are fine ignoring it?

pkg/controllers/restore/controller.go
```diff
@@ -538,7 +552,9 @@ func (h *handler) createFromDependencyGraph(ownerToDependentsList map[string][]r
 		} else {
 			resourceData = objFromBackupCR.clusterscopedResourceInfoToData[currResourceInfo]
 		}
-		if err := h.restoreResource(currResourceInfo, resourceData, objFromBackupCR.resourcesWithStatusSubresource[curr.GVR.String()]); err != nil {
+		target := fmt.Sprintf("%s.%s", currResourceInfo.GVR.Resource, currResourceInfo.GVR.GroupVersion().String())
+		hasSubStatus := slice.ContainsString(crdWithSubStatus, target)
```
nickgerace, line 556: Why isn't this an exact match over the substring search? Not opposed, just curious.

pkg/controllers/restore/controller.go
jiaqiluo: Right. We are looking for the CRD which has the field `subresources` and add it to the list `crdWithSubStatus`.

pkg/controllers/restore/controller.go
jiaqiluo: The method `slice.ContainsString` is actually doing an exact match internally:
```
func ContainsString(slice []string, item string) bool {
	for _, j := range slice {
		if j == item {
			return true
		}
	}
	return false
}
```

pkg/resourcesets/collector.go
```diff
@@ -74,18 +73,16 @@ func (h *ResourceHandler) GatherResources(ctx context.Context, resourceSelectors
 				// if this is a subresource, check if its a status subsubresource
 				if split[1] == "status" && split[0] != "customresourcedefinitions" && slice.ContainsString(res.Verbs, "update") {
```
snasovich, line 74: Previously we have skipped (`continue`) even if this check didn't succeed, now we skip only if `if` is true.

I think we should remove this `if` altogether and change comment to state that we don't gather subresources separately (note `scaling` is also a subresource).

snasovich (changes requested):

pkg/resourcesets/collector.go
snasovich: As soon as checks below result in us skipping querying `crd-name/status` and `crd-name/scaling` subresources directly then sure (do they have `list` or `get` verbs available? - I'm not sure). But it may be cleaner to just leave this check (`if len(split) == 2 {`) so it's clear we're explicitly skipping subresources.

jiaqiluo: I keep the original logic and add a debugging message.

pkg/controllers/restore/controller.go
```diff
@@ -273,20 +275,33 @@ func (h *handler) OnRestoreChange(_ string, restore *v1.Restore) (*v1.Restore, e
 	return restore, err
 }
 
-func (h *handler) restoreCRDs(created map[string]bool, objFromBackupCR ObjectsFromBackupCR) error {
+func (h *handler) restoreCRDs(created map[string]bool, objFromBackupCR ObjectsFromBackupCR) ([]string, error) {
```
snasovich, line 278: I would do a named return for `crdsWithSubStatus` here for the documentation purpose

pkg/resourcesets/collector.go
snasovich: We don't really do anything on backup when we detect status subresource, so I think it's pointless to do this check or to log this particular debug message.
I would change the whole thing to something like this (you can add a comment to explain, but I think debug message text pretty much IS that comment):
```go
if string.Contains(res.Name, "/") {
	logrus.Debugf("Skipped backing up subresource: %s", res.Name)
	continue
}
```

snasovich (changes requested):

pkg/resourcesets/collector.go
jiaqiluo: To be safe, I prefer to keep the existing check because `if string.Contains(res.Name, "/")` catches more cases than the existing logic although I am not sure if they exist.

nickgerace (dismissed):
LGTM with @snasovich's suggestions fixed

pkg/controllers/restore/controller.go
jiaqiluo: fixed.

pkg/resourcesets/collector.go
snasovich: @jiaqiluo , the current logic uses `SplitN` which returns not more than N matches. So I believe `len(strings.SplitN(res.Name, "/", 2)) == 2` is equivalent to `strings.Contains(res.Name, "/")` - I cannot think of even edge cases where these would evaluate to different values.

Also note I really want to have this long check `split[1] == "status" && split[0] != "customresourcedefinitions" && slice.ContainsString(res.Verbs, "update")` removed - our new logic is simple - if the resource we're examining is a subresource, we're skipping it.

To sum up, the suggestion is to replace current code:
```go
			split := strings.SplitN(res.Name, "/", 2)
			if len(split) == 2 {
				// if this is a subresource, check if its a status subsubresource
				if split[1] == "status" && split[0] != "customresourcedefinitions" && slice.ContainsString(res.Verbs, "update") {
					// this resource has status subresource and it accepts "update" verb, so we need to call UpdateStatus on it during restore
					logrus.Debugf("Skipped backing up subresource: %s", res.Name)
				}
				// no need to save contents of any subresource as they are a part of the resource
				continue
			}
```

with something much shorter:
```go
			if strings.Contains(res.Name, "/") {
				logrus.Debugf("Skipped backing up subresource: %s", res.Name)
				continue
			}
```

Please let me know if I'm missing something in my logic.

snasovich (changes requested):

pkg/resourcesets/collector.go
jiaqiluo: This long check `split[1] == "status" && split[0] != "customresourcedefinitions" && slice.ContainsString(res.Verbs, "update")` is what I was worried about since I wanted to avoid catching&skipping unexpected resources. 

But because it looks like there are only 2 possibilities which the res.Name containing `/`:
either `xxx/staus` or `xxx/scale`, I will take your suggested fix :) 

```
> cat test-logs.txt| grep res.Name
INFO[2021/08/23 15:41:21] res.Name: customresourcedefinitions/status
INFO[2021/08/23 15:41:22] res.Name: customresourcedefinitions/status
INFO[2021/08/23 15:41:23] res.Name: customresourcedefinitions/status
INFO[2021/08/23 15:41:23] res.Name: gitreporestrictions/status
INFO[2021/08/23 15:41:23] res.Name: clustergroups/status
INFO[2021/08/23 15:41:23] res.Name: imagescans/status
INFO[2021/08/23 15:41:23] res.Name: clusterregistrations/status
INFO[2021/08/23 15:41:24] res.Name: clusters/status
INFO[2021/08/23 15:41:24] res.Name: bundledeployments/status
INFO[2021/08/23 15:41:24] res.Name: clusterregistrationtokens/status
INFO[2021/08/23 15:41:24] res.Name: gitrepos/status
INFO[2021/08/23 15:41:24] res.Name: bundlenamespacemappings/status
INFO[2021/08/23 15:41:24] res.Name: bundles/status
INFO[2021/08/23 15:41:24] res.Name: gitjobs/status
INFO[2021/08/23 15:41:24] res.Name: customresourcedefinitions/status
INFO[2021/08/23 15:41:25] res.Name: customresourcedefinitions/status
INFO[2021/08/23 15:41:25] res.Name: clusters/status
INFO[2021/08/23 15:41:25] res.Name: vmwarevspheremachines/status
INFO[2021/08/23 15:41:26] res.Name: amazonec2machines/status
INFO[2021/08/23 15:41:26] res.Name: linodemachines/status
INFO[2021/08/23 15:41:26] res.Name: azuremachines/status
INFO[2021/08/23 15:41:26] res.Name: digitaloceanmachines/status
INFO[2021/08/23 15:41:26] res.Name: rkebootstraptemplates/status
INFO[2021/08/23 15:41:26] res.Name: rkebootstraps/status
INFO[2021/08/23 15:41:26] res.Name: custommachines/status
INFO[2021/08/23 15:41:26] res.Name: rkecontrolplanes/status
INFO[2021/08/23 15:41:26] res.Name: rkeclusters/status
INFO[2021/08/23 15:41:26] res.Name: machines/status
INFO[2021/08/23 15:41:26] res.Name: machinedeployments/status
INFO[2021/08/23 15:41:26] res.Name: machinedeployments/scale
INFO[2021/08/23 15:41:26] res.Name: clusters/status
INFO[2021/08/23 15:41:26] res.Name: machinesets/status
INFO[2021/08/23 15:41:26] res.Name: machinesets/scale
INFO[2021/08/23 15:41:26] res.Name: machinehealthchecks/status
INFO[2021/08/23 15:41:26] res.Name: projects/status
INFO[2021/08/23 15:41:27] res.Name: clusters/status
INFO[2021/08/23 15:41:27] res.Name: roletemplatebindings/status
INFO[2021/08/23 15:41:27] res.Name: roletemplates/status
INFO[2021/08/23 15:41:27] res.Name: customresourcedefinitions/status
INFO[2021/08/23 15:41:29] res.Name: customresourcedefinitions/status
INFO[2021/08/23 15:41:29] res.Name: managedcharts/status
INFO[2021/08/23 15:41:30] res.Name: apiservices/status
INFO[2021/08/23 15:41:32] res.Name: resourcesets/status
```

snasovich: @jiaqiluo , I think there is still some confusion and I would like us to be on the same page so you don't feel like I'm pushing my suggestion.

Current code skips resource as soon as there is a subresource, regardless of the result of that long `if` check. Long `if` determines whether we are adding this CRD to `statussubresource.json` and since we no longer need this, it's absolutely safe to remove this long check.

See https://github.com/rancher/backup-restore-operator/blob/e4e8c80e8f6300c42df60f9ff393132bd5069fea/pkg/resourcesets/collector.go#L72-L82, `continue` is within `if len(split) == 2` "then" block. And since we no longer need `split[i]`, my suggestion is to replace it with the simpler `Contains` check.

Are we on the same page now? I feel like I may be missing something, so really want you to confirm.

snasovich (dismissed):
LGTM

pkg/resourcesets/collector.go
jiaqiluo: Yep, we are are on the same page now 😆 

I really appreciate and like the conversation with you. Please keep doing it 💪

pkg/controllers/restore/controller.go
```diff
@@ -273,20 +275,32 @@ func (h *handler) OnRestoreChange(_ string, restore *v1.Restore) (*v1.Restore, e
...
+func (h *handler) restoreCRDs(created map[string]bool, objFromBackupCR ObjectsFromBackupCR) (crdsWithSubStatus []string, err error) {
 	for crdInfo, crdData := range objFromBackupCR.crdInfoToData {
 		err := h.restoreResource(crdInfo, crdData, false)
 		if err != nil {
-			return fmt.Errorf("restoreCRDs: %v", err)
+			return crdsWithSubStatus, fmt.Errorf("restoreCRDs: %v", err)
 		}
 		created[crdInfo.ConfigPath] = true
+
+		specs := crdData.Object[specMapKey].(map[string]interface{})
```
cbron, line 286: nit: here would be a good place for comment describing why we care about the subresource

cbron (dismissed):
lgtm, only thing I would like is comments/documentation/commit-message.

StrongMonkey (approved):
LGTM

Disclaimer: I didn't review it. Assuming this is good given other reviewers' LGTM.

PR #153, opened 2021-09-10
[2.5]Check finalizers on namespace to make sure it can be restored

StrongMonkey:
Since v1/namespace stores finalizers on spec.finalizers, we need to
check speficially if deleteTimestamp is set and finalizers are present

https://github.com/rancher/rancher/issues/34554


snasovich (approved):
LGTM

jiaqiluo (approved):
lgtm

PR #158, opened 2021-09-15
[2.6]Add ability to ignore error during restore

StrongMonkey:
https://github.com/rancher/rancher/issues/34554#issuecomment-919612531

cc @jiaqiluo @snasovich I like to run this idea to you.

### DIFF
--- a/pkg/apis/resources.cattle.io/v1/types.go
+++ b/pkg/apis/resources.cattle.io/v1/types.go
@@ -112,6 +112,9 @@ type RestoreSpec struct {
 	Prune                      *bool            `json:"prune"` //prune by default
 	DeleteTimeoutSeconds       int              `json:"deleteTimeoutSeconds,omitempty"`
 	EncryptionConfigSecretName string           `json:"encryptionConfigSecretName,omitempty"`
+
+	// When set to true, the controller ignores any errors during the restore process
+	IgnoreErrors bool `json:"ignoreErrors,omitempty"`
 }
 
 type RestoreStatus struct {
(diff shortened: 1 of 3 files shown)


jiaqiluo (approved):
LGTM

pkg/apis/resources.cattle.io/v1/types.go
```diff
@@ -112,6 +112,9 @@ type RestoreSpec struct {
 	Prune                      *bool            `json:"prune"` //prune by default
 	DeleteTimeoutSeconds       int              `json:"deleteTimeoutSeconds,omitempty"`
 	EncryptionConfigSecretName string           `json:"encryptionConfigSecretName,omitempty"`
+
+	// When set to true, controller will ignore any error during restore CRs and CRDs and mark it as done
```
snasovich, line 116: I believe we're ignoring any restore errors in this case, not just CRs and CRDs, so this comment is a bit misleading.
```suggestion
	// When set to true, the controller ignores any errors during the restore process
```

pkg/apis/resources.cattle.io/v1/types.go
```diff
@@ -112,6 +112,9 @@ type RestoreSpec struct {
...
+	IgnoreError bool `json:"ignoreError,omitempty"`
```
snasovich, line 117: Nit, but maybe have it plural since we're ignoring any number of records and not necessarily one error. Obviously, CRD yaml needs to be updated accordingly.
```suggestion
	IgnoreErrors bool `json:"ignoreErrors,omitempty"`
```

snasovich (changes requested):
Couple minor suggestions

snasovich (approved):

PR #167, opened 2021-11-01
Add FOSSA to Drone Pipeline

oxr463:
This adds a fossa analyze and test step to the drone pipeline.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -15,6 +15,27 @@ steps:
   - name: docker
     path: /var/run/docker.sock
 
+- name: fossa
+  image: registry.suse.com/suse/sle15:15.3
+  failure: ignore
+  environment:
+    FOSSA_API_KEY:
+      from_secret: FOSSA_API_KEY
+  commands:
+    - zypper -n install curl unzip
+    - "curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/spectrometer/master/install.sh | sh"
+    - fossa analyze
+    - fossa test
+  when:
+    instance:
+      - drone-publish.rancher.io
+    ref:
+      include:
+        - "refs/heads/master"
+    event:
+      - push
+      - tag
+
 - name: github_binary_release
   image: plugins/github-release
   settings:


.drone.yml
```diff
@@ -15,6 +15,27 @@ steps:
   - name: docker
     path: /var/run/docker.sock
 
+- name: fossa
+  image: registry.suse.com/suse/sle15:15.3.17.8.1
```
cbron, line 19: We've decided to just do `15:15.3` instead of the full version. I know for this case it doesn't really matter, but some of these isolated tags in the past were buggy.

.drone.yml
```diff
@@ -15,6 +15,27 @@ steps:
...
+  failure: ignore
+  environment:
+    FOSSA_API_KEY:
+      from_secret: FOSSA_API_KEY
+  commands:
+    - zypper -n install curl unzip
+    - "curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/spectrometer/master/install.sh | sh"
+    - fossa analyze
+    - fossa test
```
cbron, line 28: We should consider building a custom fossa image so that if the curl fails it doesn't kill the run, that would also speed this process up because the image will get cached on runners.

StrongMonkey: One downside of that is if we never update custom fossa image it will never pick up latest sle15 base image(15.3). Not sure if that matters

cbron: It probably doesn't because its not getting shipped with product. But this change itself is also minimal gain.

cbron (approved):

.drone.yml
StrongMonkey: Well it probably reduces the chance of getting attacked by running YOLO scripts on every build if `https://raw.githubusercontent.com/fossas/spectrometer/master/install.sh` get hacked....

cbron: @oxr463  @sirredbeard we don't necessarily have to do this, but if we are adding it everywhere then now would be the best time.

jiaqiluo (approved):
LGTM

.drone.yml
oxr463: It's pretty much already in place. I'm just adding some additional repositories.

StrongMonkey (approved):

jiaqiluo:
@oxr463 Thank you for making the PR! 
Could you backport it to the `release/v1.0` branch?  I can help review it when it is ready.

Issue #164, opened 2021-10-05
After restore fleet agent does not communicate back to fleet controller and shows error in continous delivery in Rancher 2.5.9


Martin-Weiss:
SURE-3497
SURE-3502

We are using fleet with Rancher 2.5.9, we have created several git repos that we deploy via fleet to the local and to the downstream clusters.

Now we created a rancher-backup with the backup operator, destroyed the cluster, redeployed the cluster and followed the recovery procedure described at https://rancher.com/docs/rancher/v2.5/en/backups/migrating-rancher/

After this recovery we have realized that the gitrepo and helmchartrepo secrets were not restored (see issue https://github.com/rancher/backup-restore-operator/issues/163) and we re-created this secrets manually with kubectl apply -f <secret>.yaml.

Now we had to realize that the fleet agent does not work / communicate back and complains about a missing secret fleet-system/fleet-agent-bootstrap.

We have tried to create this bootstrap secret and were somewhat successful but even with the agent working somewhat - we see it as "Cluster: local" with status red color "Modified" telling us that it does not trust the CA of the kubernetes cluster which might be the correct message as the cluster was re-deployed and we really have a new CA for the RKE2 cluster that is the basis.

Long story short - how can we restore rancher properly including all the fleet agents and the required secrets so that the agents are happy?

StrongMonkey:
Available to test with backup-operator v1.2.1-rc1. The expect result is that after backup and restore, fleet clusters should stay connect with rancher server and all clusters should be active. User should be able to use fleet after restore.

StrongMonkey:
@sowmyav27 Assigning to you right now. Depends on who you want to delegate either from red team QA or QA who has done backup testing before(@anupama2501).

sgapanovich:
Was able to reproduce issues with fleet agents and controller 
I have a 2.5.11 Rancher with 2 downstream clusters 
1. Created 3 Git repos using:
  - public repo
  - private repo
  - helm chart which needs helmsecret
2. Bundles were created and everything deployed as expected 
3. Created a backup using 1.2.0 backup
4. Migrated to a new cluster
  -"Unauthorized" errors logged in fleet agents in downstream clusters
  - `level=error msg="Failed to register agent: looking up secret fleet-system/fleet-agent-bootstrap: looking up secret fleet-system/fleet-agent-bootstrap: secrets \"fleet-agent-bootstrap\" not found"` logged in the fleet agent on my local cluster
  - ```
[...]

PR #170, opened 2021-11-10
Force sync fleet-agent bundle to remove modified states

StrongMonkey:
We need to force sync fleet-agent bundle after a restore to prevent
fleet cluster falls into modified state, as there is previously a bug
with fleet that we use two different setID to track fleet-agent bundle.
The bug is fixed in fleet v0.3.6 and above, but in order for fleet
v0.3.5 to work with backup operator we need to add this fix in.

Fix https://github.com/rancher/fleet/issues/535 in backup and restore use case.
https://github.com/rancher/backup-restore-operator/issues/164


jiaqiluo (approved):
LGTM

jiaqiluo:
Oh, @StrongMonkey Can you also forward port the PR to the master branch?

StrongMonkey:
@jiaqiluo This doesn't need to be forward-port, as I mentioned 2.6 use latest fleet version which should address the issue linked above.

PR #171, opened 2021-11-13
migrate CustomResourceDefinition's apiVersion from v1beta1 to v1

jiaqiluo:
https://github.com/rancher/rancher/issues/34154

Issue:
Warning when installing the rancher-backup-crd app
```
"apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+" 
```

the fix contains two parts:
1. migrate its own CRDs' apiVersion  to `apiextensions.k8s.io/v1`
2. update the builtin resouseSet to collect Rancher's CRDs which apiVersion  is `apiextensions.k8s.io/v1`

Fix 1 is to mitigate the warning.
Fix 2 is to make sure we can backup resources in 1.22 clusters. I have confirmed that **all**  Rancher's CRDs support the apiVersion `apiextensions.k8s.io/v1`, so no resources will be missed in the backup file with fix 2. 

Testing:
Although Rancher can not run in 1.22 cluster when I made the PR, I used the binary to make backups from and restore into a 1.22 local cluster successfully.


nickgerace (approved):
LGTM. Talked to @jiaqiluo about v1beta1 conversions and limitations. On the same boat. Approved!

pkg/crds/crd.go
```diff
@@ -68,49 +75,56 @@ func List() []crd.CRD {
...
-	schedule := spec.Properties["schedule"]
-	schedule.Description = "Cron schedule for recurring backups"
-	examples := make(map[string]interface{})
-	examples["Standard crontab specs"] = "0 0 * * *"
-	examples["Descriptors"] = "@midnight"
-	byteArr, err := json.Marshal(examples)
-	if err == nil {
-		schedule.Example = &apiext.JSON{Raw: byteArr}
+	for _, version := range backup.Spec.Versions {
+		properties := version.Schema.OpenAPIV3Schema.Properties
```
StrongMonkey, line 79: hmm so are we doing in a loop and only take the last one?

jiaqiluo: The `properties` is reset to be the current version’s during each iteration, so the changes are applied to each version. is it right?

StrongMonkey: I see the point.

StrongMonkey (approved):

PR #194, opened 2022-02-18
Add/improve logging

superseb:
- Adds logging to common operations
- Also removes the `default` make target and script because it is incomplete and confusing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,6 @@ TARGETS := $(shell ls scripts)
 $(TARGETS): .dapper
 	./.dapper $@
 
-.DEFAULT_GOAL := default
+.DEFAULT_GOAL := ci
 
 .PHONY: $(TARGETS)
--- a/pkg/resourcesets/collector.go
+++ b/pkg/resourcesets/collector.go
@@ -235,6 +235,7 @@ func (h *ResourceHandler) filterByNameAndLabel(ctx context.Context, dr dynamic.R
 	}
 	// filter out using ResourceNameRegexp
 	if filter.ResourceNameRegexp != "" {
+		logrus.Infof("Using ResourceNameRegexp [%s] to filter resource names", filter.ResourceNameRegexp)
 		if filter.ResourceNameRegexp == "." {
 			// "." will match everything, so return all resources obtained from the list call
 			return resourceObjectsList.Items, nil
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -10,6 +10,6 @@ TARGETS := $(shell ls scripts)
 $(TARGETS): .dapper
 	./.dapper $@
 
-.DEFAULT_GOAL := default
+.DEFAULT_GOAL := ci
```
eliyamlevy, line 13: Just curious, what does `ci` stand for?

pkg/resourcesets/collector.go
```diff
@@ -248,6 +249,7 @@ func (h *ResourceHandler) filterByNameAndLabel(ctx context.Context, dr dynamic.R
 				return filteredByName, err
 			}
 			if !nameMatched {
+				logrus.Debugf("Skipping %s because it did not match ResourceNameRegexp %s", name, filter.ResourceNameRegexp)
```
eliyamlevy, line 252: See comment on line 238.

pkg/resourcesets/collector.go
```diff
@@ -235,6 +235,7 @@ func (h *ResourceHandler) filterByNameAndLabel(ctx context.Context, dr dynamic.R
 	}
 	// filter out using ResourceNameRegexp
 	if filter.ResourceNameRegexp != "" {
+		logrus.Infof("Using ResourceNameRegexp %s to filter resource names", filter.ResourceNameRegexp)
```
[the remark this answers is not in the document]
prachidamle: Log statements should start with lowercase I think

pkg/resourcesets/collector.go
prachidamle: Start  Log statement with lowercase

prachidamle (changes requested):
Can we make two separate commits - one for logging and another for script removal

Makefile
superseb: Continuous Integration

pkg/resourcesets/collector.go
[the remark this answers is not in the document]
superseb: Added brackets.

Regarding log statements lowercase, nearly all log statements start with upper case and if we don't want that we need CI to prevent it.

prachidamle (approved):

Issue #165, opened 2021-10-07
After restore we see two default-tokens in some namespaces - why do we backup and restore the default token?


Martin-Weiss:
SURE-3501

Just have done a backup and restore in Rancher 2.5.9 with the backup-operator and had to realize that there are two default-tokens in all namespaces!

```
 kubectl get secret -n fleet-system
NAME                                     TYPE                                  DATA   AGE
default-token-cg6wp                      kubernetes.io/service-account-token   3      81m
default-token-zppdx                      kubernetes.io/service-account-token   3      81m
fleet-agent-token-5rr2b                  kubernetes.io/service-account-token   3      81m

```

Is this a bug or per design?

sgapanovich:
I tried to reproduce the issues with duplicate default-tokens and I couldn't 

rancher versions: 2.5.9 / 2.5.12-rc2 / 2.6-head 
backup operators: 1.2.0 (for 2.5.x) / 2.1.0 (for 2.6.x)
rancher server install: tried both docker and HA

repro steps:
1. create rancher (docker or HA)
2. create downstream clusters 
3. create users
4. create roles
5. create cluster templates
6. install back up
7. take note of the default tokens and count in the local cluster (`kubectl get secret -A | grep default- | wc -l`)
7. take back up
8. create more users, roles, cluster templates
9. restore to the same cluster
10. check that all resources created in step 8 are gone
11. check secrets and their count

didn't see extra token created

@Martin-Weiss  were you able to see the issue more than once? 
are there anything specific steps required besides described above?

Martin-Weiss:
The issue happens in case you restore to a "fresh" cluster.

sgapanovich:
Tried on 2.5.x and 2.6.x and was able to see the issue in both setups

----------

### Repro Environment: ###
Rancher version: 
  -v2.5-8fba445092b0565fc0c333d415f89596d850462b-head 
  -v2.6-e8a87f42b3a697164549a81dcc6b3a9238e5e4c4-head
Rancher cluster type: HA
Docker version: 20.10
Rancher backup version: 1.2.0 (for 2.5.x) and 2.1.0 (for 2.6.x)

----------

### Repro steps: ###
1. create a HA rke1 cluster
2. install Rancher
3. install a backup app
4. take a backup
5. create a fresh HA rke1 cluster
6. migrate your rancher to it 
  https://rancher.com/docs/rancher/v2.5/en/backups/migrating-rancher/
[...]

PR #195, opened 2022-02-18
Exclude default serviceaccount on backup

superseb:
https://github.com/rancher/backup-restore-operator/issues/165

Root cause is described in https://github.com/rancher/backup-restore-operator/issues/165#issuecomment-1045296323

### DIFF
--- a/pkg/resourcesets/collector.go
+++ b/pkg/resourcesets/collector.go
@@ -229,7 +229,7 @@ func (h *ResourceHandler) filterByNameAndLabel(ctx context.Context, dr dynamic.R
 	}
 	filteredByNameMap := make(map[*unstructured.Unstructured]bool)
 
-	if len(filter.ResourceNames) == 0 && filter.ResourceNameRegexp == "" {
+	if len(filter.ResourceNames) == 0 && filter.ResourceNameRegexp == "" && filter.ExcludeResourceNameRegexp == "" {
 		// no filters for names of the resource, return all objects obtained from the list call
 		return resourceObjectsList.Items, nil
 	}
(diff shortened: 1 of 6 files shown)


pkg/resourcesets/collector.go
```diff
@@ -255,6 +255,24 @@ func (h *ResourceHandler) filterByNameAndLabel(ctx context.Context, dr dynamic.R
 		}
 	}
 
+	// filter out using ExcludeResourceNameRegexp
+	if filter.ExcludeResourceNameRegexp != "" {
+		for _, resObj := range resourceObjectsList.Items {
+			metadata := resObj.Object["metadata"].(map[string]interface{})
+			name := metadata["name"].(string)
+			nameMatched, err := regexp.MatchString(filter.ExcludeResourceNameRegexp, name)
```
PennyScissors, line 263: If I'm understanding this correctly, this line should return `true` if the string in the variable `name` starts or ends with `default`, but I tried this in the go playground and got `false` instead. I do get that behavior if I use the pattern `^default|default$`, but I see we are using the `^<string>$` pattern to match other things already so am I missing something? https://go.dev/play/p/X1GyCu_9ymg

superseb: The exclude is for service accounts called `default`, the playground shows comparing token names with the regex. We are excluding the default service accounts created in each namespace and they are named `default`. Let me know if I need to add some clarification in the issue or code.

PennyScissors: Okay, makes sense now. I think the comment you linked to is good, this is just my first time reviewing the backup-restore-operator.

PR #202, opened 2022-03-18
Compare and update storedVersions on CRD

superseb:


pkg/util/util.go
```diff
@@ -54,3 +54,15 @@ func ErrList(e []error) error {
...
 	return nil
 }
+
+func InterfaceSlicesEqual(a, b []interface{}) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	for i, v := range a {
+		if v != b[i] {
+			return false
```
prachidamle, line 64: Will the order of the status.storedVersions in the two objects be always same? Will this matter since we are comparing items at the same index

thedadams (commented):
I have a small concern with this approach because of the way Kubernetes handles statuses.

The `storedVersions` get put on status by Kubernetes. Meaning that when the object changes, Kubernetes, presumably, has a controller that will update the status with the proper `storedVersions` slice.

So when we update the `storedVersions` here, does Kubernetes immediately update it again before the backup-restore-operator updates the spec of the CRD?

Issue #186, opened 2022-01-20
Restoring a backup results in `"backups.resources.cattle.io" is invalid: spec.preserveUnknownFields: Invalid value: true`


ADustyOldMuffin:
Full error

```
ERRO[2022/01/20 19:12:15] Error restoring CRDs restoreCRDs: restoreResource: err updating resource CustomResourceDefinition.apiextensions.k8s.io "backups.resources.cattle.io" is invalid: spec.preserveUnknownFields: Invalid value: true: cannot set to true, set x-preserve-unknown-fields to true in spec.versions[*].schema instead 
ERRO[2022/01/20 19:12:15] error syncing 'restore-from-backup': handler restore: error restoring CRDs, check logs for exact error, requeuing 
```

I believe this relates to https://github.com/kubernetes/kubernetes/issues/94295 since I know 2.1.0 recently updated to v1. Seems that field needs to be specified in the CRDs as false for v1.

## Steps to replicate
1. Install rancher backup helm chart v2.0.1
2. Take a backup
3. Upgrade helm chart to v2.1.0
4. Restore from previous backup

This occurs on any CRD that has any field changes/version changes in the lifetime of the cluster.

## Environment 
Kubernetes version: v1.21.4+rke2r2
Rancher Version: v2.6.3

thedadams:
Thanks for reporting this issue.

The difference here seems to be that the CRDs are backed up as `v1beta1` and restored as `v1`, which is causing the break. You would likely not see this issue if you backed up and restored with this same version of the backup-restore-operator (whether that is v2.0.1 or v2.1.0).

It seems that the following work-around would be available:
1. Backup with v2.0.1
2. Restore with v2.0.1
3. Upgrade to v2.1.0
4. Backup with v2.1.0
5. Restore (later) with v2.1.0

At this point, it would seem that this should be documented because of this change with CRDs from `v1beta1` to `v1`.

ajacques:
> You would likely not see this issue if you backed up and restored with this same version of the backup-restore-operator

I'm also encountering this issue. However my backup and restore operator are both at v2.1.0. I'm trying to move from one cluster (local, k8s v1.21.3+k3s1) to another (prod, k8s v1.21.9) as per the [official migration guide](https://rancher.com/docs/rancher/v2.6/en/backups/migrating-rancher/).

On my source cluster (local), it shows `v1` not `v1beta`.

```
apiVersion: apiextensions.k8s.io/v1
[...]

Issue #108, opened 2021-05-24
Add Golang unit tests


jiaqiluo:
Enhancement

We need to add tests to the program, and we can start with adding unit tests.

PR #205, opened 2022-04-08
Migrate preserveUnknownFields to be v1 valid

superseb:
https://github.com/rancher/backup-restore-operator/issues/186

### DIFF
--- a/pkg/controllers/restore/controller.go
+++ b/pkg/controllers/restore/controller.go
@@ -37,15 +37,16 @@ import (
 )
 
 const (
-	metadataMapKey     = "metadata"
-	ownerRefsMapKey    = "ownerReferences"
-	clusterScoped      = "clusterscoped"
-	namespaceScoped    = "namespaceScoped"
-	leaseName          = "restore-controller"
-	secretsMapKey      = "secrets"
-	specMapKey         = "spec"
-	subResourcesMapKey = "subresources"
-	versionMapKey      = "versions"
+	metadataMapKey           = "metadata"
+	ownerRefsMapKey          = "ownerReferences"
+	clusterScoped            = "clusterscoped"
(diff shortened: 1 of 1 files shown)


ADustyOldMuffin:
Due to the nature of the operator are you planning on adding tests?

superseb:
It seems it was pushed back multiple times already (https://github.com/rancher/backup-restore-operator/issues/108), I will put it back on the radar.

pkg/controllers/restore/controller.go
```diff
@@ -626,6 +626,24 @@ func (h *handler) restoreResource(restoreObjInfo objInfo, restoreObjData unstruc
 			}
 		}
 	}
+	// Check for invalid v1beta1 fields if the APIVersion is v1
+	if obj.GetAPIVersion() == "apiextensions.k8s.io/v1" {
+		// Invalid field is spec.preserveUnknownFields
+		if _, ok := obj.Object["spec"].(map[string]interface{})["preserveUnknownFields"]; ok {
```
prachidamle, line 632: Can we change "preserveUnknownFields" to a const string - Also if there are multiple such v1beta1 fields in the spec, it might be a good idea to store them in a map and have a utility method that checks if any of these v1beta1 fields is present in the spec fields and returns back the fields found. what do you think? T

superseb: Yes that sounds good but we don't have any other fields that we need to check or correct currently, maybe when it comes back from testing but I already ran tests for this on 2.6.0 -> 2.6.3 and did not occur any issue with this fix in place.

prachidamle: sure, we can add this in future if any more fields across new apiversions face the same problem.

prachidamle (approved):

Issue #212, opened 2022-04-21
Make imagePullPolicy configurable in chart


superseb:
Ran into this while adding integration tests. We can keep the default, but making it configurable allows for possible airgap or other restricted setups

izaac:
According to upstream helm and kubernetes shouldn't the policy be:

`Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'`

https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting

superseb:
We can't just change the default, it was implemented as `Always` as default and now that we made it configurable, we cannot just change it as the user has an expectation that it is `Always`.

izaac:
@superseb ok I understand that's how it's designed but I still think we should add `imagePullPolicy: Always` explicitly to make it visible during the chart install.

izaac:
Rancher version: `v2.6.5-rc9`
Rancher backups version: `v2.1.2-rc4`
local RKE1 cluster.


I was able to see the `imagePullPolicy` option from the `values.yaml` UI component.

Default as expected `Always`
I was able to also install the chart it using `imagePullPolicy: IfNotPresent` and `imagePullPolicy: Never` and the policy was applied to the Deployment.

Updates to the chart changing the value of `imagePullPolicy` also worked.

Sample install UI `value.yaml` displaying the `imagePullPolicy`:

![Screen Shot 2022-05-10 at 12 14 19 PM](https://user-images.githubusercontent.com/27659/167706144-6e3eaf7d-3177-47de-8829-b30f9356c2af.png)

PR #213, opened 2022-04-21
Make imagePullPolicy configurable in chart

superseb:
https://github.com/rancher/backup-restore-operator/issues/212

Similarly implemented as in `rancher/rancher`: https://github.com/rancher/rancher/blob/release/v2.6/chart/templates/deployment.yaml#L67


eliyamlevy (approved):
Looks good, as we discussed this adds support for the integration tools we want to add and keeps the original functionality.

Issue #219, opened 2022-04-25
Set logrus Formatter as early as possible


superseb:
In v2.1.2-rc2, the log lines are:

```
$ kubectl  -n cattle-resources-system logs -l  app.kubernetes.io/name=rancher-backup   -f
time="2022-04-25T14:11:28Z" level=debug msg="Loglevel set to [debug]"
time="2022-04-25T14:11:28Z" level=info msg="Starting backup-restore controller version v0.0.0-dev (HEAD)"
INFO[2022/04/25 14:11:28] No PVC or S3 details provided for storing backups by default. User must specify storageLocation on each Backup CR 
INFO[2022/04/25 14:11:28] Secrets containing encryption config files must be stored in the namespace cattle-resources-system 
INFO[2022/04/25 14:11:28] Starting resources.cattle.io/v1, Kind=Backup controller 
INFO[2022/04/25 14:11:28] Starting resources.cattle.io/v1, Kind=Restore controller 
```

This shows different formatting and the reason is because formatting only happens after the first two lines.

izaac:
Rancher version: `v2.6.5-rc8`
Rancher backups chart version: `v2.1.2-rc3`


I've checked the INFO, DEBUG, TRACE log levels and the logs have only one formatting now

```yaml
 image: rancher/backup-restore-operator:v2.1.2-rc3
 args: ["-debug"]
```

Sample log with TRACE:

```
kubectl  -n cattle-resources-system logs -l  app.kubernetes.io/name=rancher-backup   -f
TRAC[2022/05/09 16:54:17] Loglevel set to [trace]
INFO[2022/05/09 16:54:17] Starting backup-restore controller version v0.0.0-dev (HEAD)
INFO[2022/05/09 16:54:17] No PVC or S3 details provided for storing backups by default. User must specify storageLocation on each Backup CR
INFO[2022/05/09 16:54:17] Secrets containing encryption config files must be stored in the namespace cattle-resources-system
INFO[2022/05/09 16:54:17] Starting resources.cattle.io/v1, Kind=Restore controller
INFO[2022/05/09 16:54:17] Starting resources.cattle.io/v1, Kind=Backup controller
```

PR #220, opened 2022-04-25
Set log formatter as early as possible

superseb:
https://github.com/rancher/backup-restore-operator/issues/219


eliyamlevy (approved):
Looks good

Issue #217, opened 2022-04-25
Make debug and trace loglevels configurable in chart


superseb:
Debug and trace flags were added in https://github.com/rancher/backup-restore-operator/pull/204/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R61, should be configurable in chart.

Similar to `rancher/rancher`: https://github.com/rancher/rancher/blob/v2.6.4/chart/values.yaml#L35

eliyamlevy:
PR: https://github.com/rancher/backup-restore-operator/pull/227

PR #227, opened 2022-05-04
Make log levels configurable

eliyamlevy:
PR for issue https://github.com/rancher/backup-restore-operator/issues/217

### DIFF
--- a/charts/rancher-backup/README.md
+++ b/charts/rancher-backup/README.md
@@ -42,6 +42,8 @@ The following table lists the configurable parameters of the rancher-backup char
 | persistence.storageClass |  StorageClass to use for dynamically provisioning the Persistent Volume, which will be used for storing backups | "" |
 | persistence.volumeName |  Persistent Volume to use for storing backups | "" |
 | persistence.size |  Requested size of the Persistent Volume (Applicable when using dynamic provisioning) | "" |
+| debug | Set debug flag for backup-restore deployment | false |
+| trace | Set trace flag for backup-restore deployment | false |
 | nodeSelector | https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector | {} |
 | tolerations | https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration | [] |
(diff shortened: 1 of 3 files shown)


charts/rancher-backup/templates/deployment.yaml
```diff
@@ -65,4 +72,4 @@ spec:
       tolerations: {{ include "linux-node-tolerations" . | nindent 8 }}
 {{- if .Values.tolerations }}
 {{ toYaml .Values.tolerations | indent 8 }}
-{{- end }}
+{{- end }}
```
superseb, line 75: We probably need to decide wether we want newlines at end of file or not, this is usually a code editor setting and if we don't decide what we do, we keep getting these changes over and over. I've seen more PRs with a newline than without.

superseb, line 75: We probably need to decide wether we want newlines at end of file or not, this is usually a code editor setting and if we don't decide what we do, we keep getting these changes over and over. I've seen more PRs with a newline than without.

superseb:
This also needs a change in https://github.com/rancher/backup-restore-operator/blob/master/charts/rancher-backup/README.md to describe configurable options

superseb:
This also needs a change in https://github.com/rancher/backup-restore-operator/blob/master/charts/rancher-backup/README.md to describe configurable options

superseb:
And please squash commits to a single one

superseb:
And please squash commits to a single one

charts/rancher-backup/README.md
```diff
@@ -42,6 +42,8 @@ The following table lists the configurable parameters of the rancher-backup char
 | persistence.storageClass |  StorageClass to use for dynamically provisioning the Persistent Volume, which will be used for storing backups | "" |
 | persistence.volumeName |  Persistent Volume to use for storing backups | "" |
 | persistence.size |  Requested size of the Persistent Volume (Applicable when using dynamic provisioning) | "" |
+| debug | Set debug flag for backup-restore deployment | false |
```
superseb, line 45: Can we change this to `Enable debug logging for operator` and `Enable trace logging for operator (this can contain sensitive info)`

superseb (approved):

PR #251, opened 2022-06-22
Update chart version constraints

superseb:
See https://github.com/rancher/charts/pull/1948#discussion_r903121177 and https://github.com/rancher/charts/pull/1948#discussion_r903121442

Also updated `kubeVersion` to match `catalog.cattle.io/kube-version`

### DIFF
--- a/charts/rancher-backup/Chart.yaml
+++ b/charts/rancher-backup/Chart.yaml
@@ -2,12 +2,12 @@ annotations:
   catalog.cattle.io/auto-install: rancher-backup-crd=match
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: Rancher Backups
-  catalog.cattle.io/kube-version: '>= 1.16.0-0'
+  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.25.0-0'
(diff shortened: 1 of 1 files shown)


charts/rancher-backup/Chart.yaml
```diff
@@ -21,6 +21,6 @@ icon: https://charts.rancher.io/assets/logos/backup-restore.svg
 keywords:
 - applications
 - infrastructure
-kubeVersion: '>= 1.16.0-0'
+kubeVersion: '>= 1.16.0-0 < 1.25.0-0'
```
PennyScissors, line 24: We should leave this one as is without the max. This change would just introduce another value we would need to maintain and if we ever miss it, changing the kubeVersion on a released asset is not as easy like we change the kube-version annotation because it breaks stuff. So is better to leave it open ended or not use it at all. @aiyengar2 Please correct me if I'm wrong.

superseb: 😛 The changes in this PR come directly from your review, and this one comes from this comment specifically https://github.com/rancher/charts/pull/1948#discussion_r903121177

PennyScissors (approved):

Issue #258, opened 2022-07-26
Binary backup-restore-operator v2.1.3-rc4 is built for x86_64 only


thehejik:
When deploying `Rancher Backups v2.1.3-rc4` App in `Rancher:v2.6.7-rc3` for `s390x` I noticed that the deployment is failing. 

Culprit is that image `rancher/backup-restore-operator:v2.1.3-rc4` for `s390x` contains binary `/usr/bin/backup-restore-operator` which was built for `x86_64`.

When looking into GH Releases/Assets I found out that the size and checksums are the same for all platforms incl. `aarch64`.

Also checked by `file` util: 
```
> file backup-restore-operator-s390x
backup-restore-operator-s390x: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, stripped
```

superseb:
Verified in https://github.com/rancher/backup-restore-operator/pull/259#pullrequestreview-1050720395

PR #259, opened 2022-07-26
Remove hardcoded arch for build

superseb:
https://github.com/rancher/backup-restore-operator/issues/258


thehejik (approved):
Verified on my s390x vm that `make` produces a valid s390x binary

PR #267, opened 2022-09-02
Allow static pv to be correctly referenced.

### LINKED ISSUE #94: Make PVC name customisable in backup-restore-operator
Using backup-restore operator with a storage class for persistence of snapshots results in a pvc claim being suffixed with the Helm ReleaseRevision.

This means each time an update to the backup-restore-operator is rolled out it results in a new pvc claim being created.

```
{{/*
Create PVC name using release and revision number.
*/}}
{{- define "backupRestore.pvcName" -}}
{{- printf "%s-%d" .Release.Name .Release.Revision }}
{{- end }} 
```

We should either re-use pvc name or allow the users to customize the PVC name so they can reuse existing PV's.

RudiMT:
Fix #94. The generated pvc name and the size must be passed as well for this to work

### DIFF
--- a/charts/rancher-backup/templates/_helpers.tpl
+++ b/charts/rancher-backup/templates/_helpers.tpl
@@ -79,7 +79,7 @@ Create PVC name using release and revision number, unless a volumeName is given.
 */}}
 {{- define "backupRestore.pvcName" -}}
 {{- if and .Values.persistence.volumeName }}
-{{- printf "%s-pvc" .Values.persistence.volumeName }}
+{{- printf "%s" .Values.persistence.volumeName }}
 {{- else -}}
 {{- printf "%s-%d" .Release.Name .Release.Revision }}
 {{- end }}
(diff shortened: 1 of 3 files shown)


superseb (commented):
Sorry for the late review @RudiMT. I think it's a simple change and should work fine, the only question I have if it someone wanted to customize the name, they are still bound to the static `-pvc` at the end which might not work for someone's naming convention or something?

RudiMT:
> Sorry for the late review @RudiMT. I think it's a simple change and should work fine, the only question I have if it someone wanted to customize the name, they are still bound to the static `-pvc` at the end which might not work for someone's naming convention or something?

I'm fine with -pvc and without. The PVC name isn't a parameter currently, though, so there's no possibility for the users to use their own convention (and I didn't want to add that logic). The docs say "persistence.enabled: Configure a Persistent Volume as the default storage location. It accepts either a StorageClass name to create a PVC, or directly accepts the PV to use.", but don't say how the PVC's name is derived in either case.

superseb:
@RudiMT Thanks! Please squash commits to one

superseb (changes requested):
^^

RudiMT (commented):
OK

superseb:
Superseded by https://github.com/rancher/backup-restore-operator/pull/274

PR #287, opened 2022-12-16
Created 1.25 test pipeline, added capability to set k8s version in pipeline

eliyamlevy:


Dockerfile.dapper
```diff
@@ -22,8 +22,17 @@ ENV K3S_BINARY_amd64=k3s \
...
-RUN curl -sL https://github.com/rancher/k3s/releases/download/$(curl -Ls -o /dev/null -w %{url_effective} https://update.k3s.io/v1-release/channels/v1.24 | awk -F/ '{ print $NF }')/${!K3S_BINARY} > /usr/local/bin/k3s && \
-    chmod +x /usr/local/bin/k3s
+ARG PSP_ENABLED
+
+RUN echo "${PSP_ENABLED}"
+
+RUN if [ "${PSP_ENABLED}" == "1" ]; then \
+        curl -sL https://github.com/rancher/k3s/releases/download/$(curl -Ls -o /dev/null -w %{url_effective} https://update.k3s.io/v1-release/channels/v1.24 | awk -F/ '{ print $NF }')/${!K3S_BINARY} > /usr/local/bin/k3s; \
+    else \
+        curl -sL https://github.com/rancher/k3s/releases/download/$(curl -Ls -o /dev/null -w %{url_effective} https://update.k3s.io/v1-release/channels/stable | awk -F/ '{ print $NF }')/${!K3S_BINARY} > /usr/local/bin/k3s; \
```
stormqueen1990, line 32: This block is largely the same for both if and else. How do you feel about using a variable for the version, then extracting the `curl`s from if/else?

stormqueen1990 (commented):
How do you feel about splitting the changes to the chart from the changes to the pipeline? It seems they're not directly related.

Dockerfile.dapper
eliyamlevy: Hmm, that might look cleaner. I'll redo the pr

eliyamlevy: Actually I just remembered, the whole point of that was for people building locally so theres a default value. Either way its gonna be a bit clunky, whether we repeat that line or have a whole if else block just to set the value.

Dockerfile.dapper
```diff
@@ -22,7 +22,15 @@ ENV K3S_BINARY_amd64=k3s \
     K3S_BINARY_s390x=k3s-s390x \
     K3S_BINARY=K3S_BINARY_${ARCH}
 
-RUN curl -sL https://github.com/rancher/k3s/releases/download/$(curl -Ls -o /dev/null -w %{url_effective} https://update.k3s.io/v1-release/channels/v1.24 | awk -F/ '{ print $NF }')/${!K3S_BINARY} > /usr/local/bin/k3s && \
+ARG K8S_VERSION_FROM_DRONE
+ENV K8S_VERSION $K8S_VERSION_FROM_DRONE
+
+RUN if [ -z "${K8S_VERSION}" ]; then export K8S_VERSION="v1.24" && echo $(date +%s%N); fi
+# ENV K8S_VERSION="${K8S_VERSION_FROM_DRONE:-'v1.24'}"
```
stormqueen1990, line 29: nit: this could go away.

Dockerfile.dapper
```diff
@@ -22,7 +22,15 @@ ENV K3S_BINARY_amd64=k3s \
...
+
+RUN echo "${K8S_VERSION}"
+
+RUN curl -sL https://github.com/rancher/k3s/releases/download/$(curl -Ls -o /dev/null -w %{url_effective} https://update.k3s.io/v1-release/channels/${K8S_VERSION} | awk -F/ '{ print $NF }')/${!K3S_BINARY} > /usr/local/bin/k3s && \
```
stormqueen1990, line 33: It's good practice to enclose in double quotes strings that use interpolation in bash (to prevent word and glob splitting, if the expansion evaluates to values that contain spaces).

stormqueen1990 (approved):
LGTM. I left a comment on enclosing variables in double quotes. I see there are more occurrences so this could be addressed in a separate PR.

Issue #163, opened 2021-10-05
Backup/Restore does not include the secrets used for clientSecretName and helmSecretName in fleet gitrepos


Martin-Weiss:
SURE-3503
SURE-3531 - closed as duplicate

After a rancher restore in Rancher 2.5.9 the gitrepos defined in fleet do not work anymore as we have to use authentication against git and against the helm repo (harbor in our case) and the secrets  defined in clientSecretName and helmSecretName of the gitrepos defined are not backed up / restored.

Could we get the backup operator fixed in a way that it automatically creates a backup of these secrets in the fleet-default / fleet-local namespaces, too? (maybe also in other fleet namespaces if additional ones have been created)

MKlimuszka:
@nickwsuse , Eli believes this is already fixed.

prachidamle:
Could we retest this usecase with latest Rancher version and latest Backup-restore operator to see if the issue is still seen? There is a high chance that latest versions might have this resolved already

nickwsuse:
@eliyamlevy @prachidamle I'm not seeing the secrets I made before taking the backup after I deleted them and restored. The secrets I made are opaque, so not sure if there's an expected secret type for this to work, but I would assume it should keep all secrets in the namespaces regardless of type?

Steps:
1. Created a v2.7-head instance (Commit ID: 65089ee)
2. Created a downstream RKE1 and RKE2 cluster
3. Added an opaque secret with my S3 creds
4. Installed Rancher Backups v3.0.1-rc2 with my AWS S3 bucket info
5. Created a backup called "pre-fleet-secrets"
6. Created an opaque secret in the fleet-default namespace
7. Created an opaque secret in the fleet-local namespace
8. Created another backup called "post-fleet-secrets"
9. Deleted the secrets I made in both fleet namespaces
10. Created a restore using the "post-fleet-secrets" backup
11. Checked for the secrets I made that should have been restored

This test was a local restore, but I imagine we'd get the same results with a restore done in a migration as well.

MKlimuszka:
@eliyamlevy , we checked with Ron and he said it would be ok to add this to the Q1 release.

eliyamlevy:
Hello all,
The solution we came to was to create a resourceSelector which will look for any secret with a specific label to back it up. This selector will look in all namespaces and at every secret on the cluster. If the secret has the label `resources.cattle.io/backup: true` it will be backed up.  

[...]

PR #291, opened 2023-01-10
Added resourceSelector for secrets with resources.cattle.io/backup label

eliyamlevy:
PR for [163](https://github.com/rancher/backup-restore-operator/issues/163), the solution is to create a specific label to assign to secrets, which will be searched for using a resource selector added to the default resource set.


charts/rancher-backup/files/default-resourceset-contents/rancher.yaml
```diff
@@ -52,3 +52,11 @@
   kindsRegexp: "^clusterrepos$"
 - apiVersion: "resources.cattle.io/v1"
   kindsRegexp: "^ResourceSet$"
+- apiVersion: "v1"
+  kindsRegexp: "^secrets$"
+  namespaceRegexp: "."
```
aiyengar2, line 57: should this be `^.*$`

eliyamlevy: I believe you're right, I was following the other regex in the resourceSet that used a `.` to capture all kinds.

eliyamlevy: Made your change and tested, found four test secrets in four different namespaces in the backup. So looks like it works!

aiyengar2 (approved):

PR #297, opened 2023-01-27
Updated psp config to current rancher approach

eliyamlevy:
[issue](https://github.com/rancher/rancher/issues/40348)


stormqueen1990 (commented):
Changes LGTM.

You may want to add a blurb to the main chart README.md explainining the change in behaviour between before and after the change (e.g. before PSPs were default when available but now they're not going to be installed unless the user explicitly asks for it).

eliyamlevy:
> Changes LGTM.
> 
> You may want to add a blurb to the main chart README.md explainining the change in behaviour between before and after this PR (e.g. before PSPs were default when available but now they're not going to be installed unless the user explicitly asks for it).

@stormqueen1990  how about just just adding it to the configuration settings section?

stormqueen1990 (approved):

PR #310, opened 2023-03-15
Change restore controller to not restore secrets with type fleet.cattle.io/cluster-registration-values

eliyamlevy:
## Issue: 

[Parent Issue](https://github.com/rancher/rancher/issues/40819#issuecomment-1468969549)
 
## Problem
 Created as a result of an issue with Rancher where on migration from one installation to another, Fleet is updating the agent on a downstream cluster with invalid fleet-agent-bootstrap credentials. 

Additional context: https://github.com/rancher/rancher/issues/40819#issuecomment-1468969549

From resulting discussions with @snasovich and @Oats87 we decided to not restore secrets with type [fleet.cattle.io/cluster-registration-values]. The following solution is temporary until a more concrete feature will be added.

## Solution

Added below to `pkg/controllers/restore/controller.go` so no secrets with type [fleet.cattle.io/cluster-registration-values](http://fleet.cattle.io/cluster-registration-values)  are restored.
```
if gvr.Resource == "secrets" {
	secretType, found, err := unstructured.NestedString(obj.Object, "type")
	if err != nil {
			return err
	}
	if found && secretType == "fleet.cattle.io/cluster-registration-values" {
		logrus.Infof("Skip restoring the secret %s/%s since it has type fleet.cattle.io/cluster-registration-values", namespace, name)
		return nil
	}
}
```
 
## Testing
 Repro steps in parent issue are valid.

### DIFF
--- a/pkg/controllers/restore/controller.go
+++ b/pkg/controllers/restore/controller.go
@@ -606,6 +606,17 @@ func (h *handler) restoreResource(restoreObjInfo objInfo, restoreObjData unstruc
 	name := restoreObjInfo.Name
 	namespace := restoreObjInfo.Namespace
 	gvr := restoreObjInfo.GVR
+	// TEMPORARY HOTFIX: Don't restore secrets of type fleet.cattle.io/cluster-registration-values
+	if gvr.Resource == "secrets" {
+		secretType, found, err := unstructured.NestedString(obj.Object, "type")
+		if err != nil {
+			return err
+		}
+		if found && secretType == "fleet.cattle.io/cluster-registration-values" {
+			logrus.Infof("restoreResource: Skiping secret %s/%s since it has type fleet.cattle.io/cluster-registration-values", namespace, name)
+			return nil
+		}
+	}
 	var dr dynamic.ResourceInterface
 	dr = h.dynamicClient.Resource(gvr)
 	if namespace != "" {


nflynt (approved):
Does what it says on the tin, I tested this behavior in isolation. The key is still backed up, but ignored during restore. We'll need to cross-check this against https://github.com/rancher/rancher/issues/40819#issuecomment-1468969549 and make sure it's sufficient to resolve the issue.

pkg/controllers/restore/controller.go
```diff
@@ -606,6 +606,17 @@ func (h *handler) restoreResource(restoreObjInfo objInfo, restoreObjData unstruc
...
 	namespace := restoreObjInfo.Namespace
 	gvr := restoreObjInfo.GVR
+	// TEMPORARY HOTFIX: Don't restore secrets of type fleet.cattle.io/cluster-registration-values
+	if gvr.Resource == "secrets" {
+		secretType, found, err := unstructured.NestedString(obj.Object, "type")
+		if err != nil {
+			return err
+		}
+		if found && secretType == "fleet.cattle.io/cluster-registration-values" {
+			logrus.Infof("Skip restoring the secret %s/%s since it has type fleet.cattle.io/cluster-registration-values", namespace, name)
```
snasovich, line 616: Suggesting to prefix message with `restoreResource: ` as done in all other log messages across this func.
Another nit is going with `Skipping` over `Skip` to align with format of other log messages (may be wrong on this one since it follows with `restoring`).

eliyamlevy: I'll fix this

nflynt (approved):

PR #311, opened 2023-03-20
Exclude import-token secrets from fleet backups

nflynt:
Pertaining to the following github issues:

https://github.com/rancher/rancher/issues/40819
https://github.com/rancher/rancher/issues/40080
https://github.com/rancher/rancher/issues/40859

The operator has a temporary hotfix to ignore these keys upon restore. We should avoid backing them up in the first place.


eliyamlevy (approved):
LGTM

PR #312, opened 2023-03-20
Add priorityclasses for fleet in a list of resources to backup/restore

maxsokolovsky:
Issue: https://github.com/rancher/rancher/issues/40885


charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
```diff
@@ -19,6 +19,9 @@
 - apiVersion: "v1"
   kindsRegexp: "^configmaps$"
   namespaceRegexp: "^cattle-fleet-|^fleet-|^cluster-fleet-"
+- apiVersion: "scheduling.k8s.io/v1"
+  kindsRegexp: "^priorityclasses$"
+  namespaceRegexp: "^cattle-fleet-|^fleet-|^cluster-fleet-"
```
MbolotSuse, line 24: If my understanding is correct, priority classes are not namespaced, so this line does not help us here.

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
```diff
@@ -19,6 +19,9 @@
 - apiVersion: "v1"
   kindsRegexp: "^configmaps$"
   namespaceRegexp: "^cattle-fleet-|^fleet-|^cluster-fleet-"
+- apiVersion: "scheduling.k8s.io/v1"
```
MbolotSuse, line 22: Right now this will get all priorityClasses - we only want to backup the one that the rancher chart specifies. You can do this with a `resourceNameRegexp` field, like [here](https://github.com/rancher/backup-restore-operator/blob/7e8dc2f00dad7879160be4f0bff2ec55b27fb215/charts/rancher-backup/files/default-resourceset-contents/fleet.yaml#L3). 

I would recommend anchoring the start and the end so that we exactly match what [rancher's chart](https://github.com/rancher/rancher/blob/a841a9e101ae601a033fa8b39342ad4800c91d85/chart/templates/priorityClass.yaml#L4) defines  - so something like `resourceNameRegexp: "^rancher-critical$"`

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
MbolotSuse, line 1: This priority class does not belong to fleet. Rather, it belongs to rancher's chart and is used by fleet. IMO, this more properly belongs in the [rancher](https://github.com/rancher/backup-restore-operator/blob/release/v3.0/charts/rancher-backup/files/default-resourceset-contents/rancher.yaml) resource set.

MbolotSuse (changes requested):
A couple of comments on structure and implementation.

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
maxsokolovsky: Agreed!

MbolotSuse (approved):
One nit, but LGTM (up to you on if you want to fix the nit). Please wait for a review from team 3 before merging.

eliyamlevy (approved):

Issue #306, opened 2023-02-23
[2.7] Update Wrangler Go Package


eliyamlevy:
**Rancher Server Setup**
- Rancher version: `v2.7`

**Describe the bug**
We need to update to wrangler `v1.0.2` for a cve patch. This is for `v2.7` so must go into `release/v3.0`.

## Testing

Since this is just updating packages/modules/images the general functionality should be tested. However there should be a specific round of testing focused on the encryption feature. Since this wrangler update led to having to update how the encryption transformer overrides are loaded.

eliyamlevy:
It seems that wrangler `v1.x`  line uses the k8s `v1.25` client. This means until we create a strictly `v1.25` release line for backup-restore-operator, it cannot be updated past the `v0.8.x` line of wrangler.

prachidamle:
@eliyamlevy Can this ticket be moved to-test? Can we add tests for this ticket? Also what testcases should QA ensure to cover?

eliyamlevy:
@prachidamle Just cut new RC to test a wave of updated packages and consumed upstream elements. Working on charts pr now.

eliyamlevy:
New pr open

ronhorton:
**Pass** verified in rancher `v2.7-head commit iD b5627bd`
1. installed rancher-backups
2. created backup with encryption
3. backup completed
4. restored successfully from backup taken in step 3

PR #340, opened 2023-04-25
Update wrangler to v1.0.2

eliyamlevy:
https://github.com/rancher/backup-restore-operator/issues/306


pkg/util/util.go
```diff
@@ -35,7 +37,15 @@ func GetEncryptionTransformers(encryptionConfigSecretName string, secrets v1core
 	if !ok {
 		return transformerMap, fmt.Errorf("no encryptionConfig provided")
 	}
-	return encryptionconfig.ParseEncryptionConfiguration(bytes.NewReader(encryptionConfigBytes))
+	err = ioutil.WriteFile("encryptionConfig.yaml", encryptionConfigBytes, os.ModePerm)
```
jiaqiluo, line 40: I'd suggest adding a `defer` block to delete this file.

jiaqiluo (changes requested):

pkg/util/util.go
eliyamlevy: True!

pkg/util/util.go
```diff
@@ -35,7 +37,16 @@ func GetEncryptionTransformers(encryptionConfigSecretName string, secrets v1core
...
```
jiaqiluo, line 40: Nit: declare the filename as a const at the top of the file instead of a hardcoding value in the code.

jiaqiluo (approved):
LGTM with one nit.

jiaqiluo (approved):

PR #322, opened 2023-04-17
Adding Hull Integration to Backup-Restore

eliyamlevy:

PR #347, opened 2023-05-01
Fixed helper func to grab version from env

eliyamlevy:


nflynt (approved):
Looks better than the last time I said LGTM, now with a stronger understanding of the drone environment. But we won't know if it works works until it runs.

eliyamlevy:
https://github.com/rancher/backup-restore-operator/pull/322

PR #357, opened 2023-05-12
Fixing encryption transformer override config error

eliyamlevy:
https://github.com/rancher/backup-restore-operator/issues/306


nflynt (approved):
Yeah we should probably actually use the configured filename. LGTM

PR #401, opened 2024-01-23
Changed filter logic for include and exclude resource

inichols:
Moved ExcludeResourceNameRegexp check in the same if block as ResourceNameRegexp This is to make it so ExcludeResourceNameRegexp does not causes previous check to be ignored

### DIFF
--- a/pkg/resourcesets/collector.go
+++ b/pkg/resourcesets/collector.go
@@ -235,24 +235,22 @@ func (h *ResourceHandler) filterByNameAndLabel(ctx context.Context, dr dynamic.R
 		// no filters for names of the resource, return all objects obtained from the list call
 		return resourceObjectsList.Items, nil
 	}
+
 	// filter out using ResourceNameRegexp
 	if filter.ResourceNameRegexp != "" {
 		logrus.Debugf("Using ResourceNameRegexp [%s] to filter resource names", filter.ResourceNameRegexp)
-		if filter.ResourceNameRegexp == "." {
-			// "." will match everything, so return all resources obtained from the list call
-			return resourceObjectsList.Items, nil
-		}
-
 		for _, resObj := range resourceObjectsList.Items {
-			metadata := resObj.Object["metadata"].(map[string]interface{})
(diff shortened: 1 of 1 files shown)


pkg/resourcesets/collector.go
```diff
@@ -235,7 +235,7 @@ func (h *ResourceHandler) filterByNameAndLabel(ctx context.Context, dr dynamic.R
 		// no filters for names of the resource, return all objects obtained from the list call
 		return resourceObjectsList.Items, nil
 	}
-	// filter out using ResourceNameRegexp
+	// filter out using ResourceNameRegexp and ExcludeResourceNameRegexp
 	if filter.ResourceNameRegexp != "" {
```
eliyamlevy, line 239: The problem with this solution is that if `ResourceNameRegexp` is empty (`""`) then it will never hit the ExcludeResourceNameRegexp logic. I think the logic for exclude needs to be separate.

pkg/resourcesets/collector.go
```diff
@@ -235,24 +235,22 @@ func (h *ResourceHandler) filterByNameAndLabel(ctx context.Context, dr dynamic.R
 		// no filters for names of the resource, return all objects obtained from the list call
 		return resourceObjectsList.Items, nil
 	}
+
 	// filter out using ResourceNameRegexp
 	if filter.ResourceNameRegexp != "" {
 		logrus.Debugf("Using ResourceNameRegexp [%s] to filter resource names", filter.ResourceNameRegexp)
-		if filter.ResourceNameRegexp == "." {
```
mallardduck, line 241: This is a different bug unrelated to the one we were trying to fix - HOWEVER, fixing this bug was a happy accident along the way. It helps ensure that our other fixes (which in general make things much less greedy) don't cause other unexpected bugs.

mallardduck (approved):

eliyamlevy (approved):
LGTM

ericpromislow (approved):
LGTM

PR #406, opened 2024-01-25
Small Collector Refactor & Add initial Unit tests

mallardduck:
This is just a Proof of Concept to noodle on some ideas to improve testing.

These changes fully separates the concerns of fetching API data and filtering said data. In turn this allows the filtering logic to be unit tested outside of a live cluster. And those unit tests will ensure greater stability in the future by failing "loud and early" as compared to the current testing.

It goes without saying that these functional unit tests would be additional tests on top of controller and integration testing with the wider Rancher systems.

### DIFF
--- a/pkg/resourcesets/collector.go
+++ b/pkg/resourcesets/collector.go
@@ -193,7 +193,14 @@ func (h *ResourceHandler) gatherObjectsForResource(ctx context.Context, res k8sv
 	dr = h.DynamicClient.Resource(gvr)
 
 	// only resources that match name+namespace+label combination will be backed up, so we can filter in any order
-	filteredByName, err := h.filterByNameAndLabel(ctx, dr, filter)
+	// however, in practice filtering by label happens at an API level when we paginate the resources (creating our initial list)
+	filteredByLabel, err := h.filterByLabel(ctx, dr, filter)
+	if err != nil {
+		return filteredObjects, err
+	}
+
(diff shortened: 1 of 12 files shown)


mallardduck:
Also I know this PR is super rough and the stubs should live somewhere else. Just created this mostly as a proof of concept to noodle on the idea.

ericpromislow:
Hi, can you ping me on slack?

mallardduck:
As of the newest commit on this PR the only failing test is the new one added for Namespace filtering.
However with namespaces, we only need to consider includes and not excludes; so the logic is different than name filtering.

That might warrant removing that test and adding that into another PR/branch.

So far these new unit tests and changes would cover a good chunk of the lines in the collector. And the changes being made seek to make it conform to the explanation here:
https://github.com/rancher/backup-restore-operator/blob/2d3421efda1032fcb2d14fd8853e104420f2f6f2/pkg/resourcesets/collector.go#L40-L47

While these changes will produce a difference in the backup as with the previous PR (https://github.com/rancher/backup-restore-operator/pull/401) - they seek to fix bugs and make the behaviour conform to previous expectations. That includes preserving functionality for all existing filter options (and respective combinations); as to not break potential unexpected customer use-cases that are currently valid.

ericpromislow (approved):
Good set of unit tests here, and clearer code.

Issue #411, opened 2024-01-30
[CodeQuality] Fix filtering on object's Kind or Name 


ericpromislow:
**Describe the bug**
Follow-up to bug BRO #408 : some objects were being stored multiple times, or backed up when they shouldn't have been.

On analysis, the code was not only sometimes incorrect, but the logic was unnecessarily complicated

**Plan**

Most likely we can just cherry-pick the relevant commits from 4.0 to 5.0

MKlimuszka:
https://github.com/mallardduck/backup-restore-operator/pull/1

ericpromislow:
Done as part of resourceset collector PR #415

MKlimuszka:
@ericpromislow , verify which branches these changes have gone into.

ericpromislow:
The changes are in branches `release/v3.0` - commit 0c9956568791501cd186b0b41ca94280c9dd1cfd (last commit)
and `release/v4.0` - commit 18086f5c3cbd577c3d49ecfbcec94c97b13bda50

mallardduck:
As of today these changes should be in all branches uniformly all the way up to 5.0.
Additionally they have been released for:

| Rancher | BRO |
| --- | --- |
| 2.7 | 3.1.3 |
| 2.8 | 4.0.1 |
| 2.9 | 5.0.0-rc3 (TBD) |

I've included 2.9/5.x since as of merge of https://github.com/rancher/backup-restore-operator/pull/432 the changes will be included with the next RC we cut which will be rc3.

PR #412, opened 2024-01-30
Refactor some of the resource-filterers

ericpromislow:
Fixes issue #411 

The `ResourceName`, `Namespace` and `Kind` filters all seemed somewhat clunky,  not to mention that 
they were the source of a recent bug. This PR changes them to standard filters, where you loop through the
array of source-objects and return an array of acceptable objects. 

I also precompile non-empty regular expressions, so we don't need to check for an error condition while
doing pattern-matching, making the actual filters follow a functional-programming paradigm, which in general
filters should try for.

Benchmarks showed a slight increase in performance. Running unit tests with `-count 10000`, the 
before run took 19.0 sec, the after 17.7. Note that this benchmark was done before the `TestFilterByKind_CheckDuplicates` was added, so comparisons are no longer valid.

### DIFF
--- a/pkg/resourcesets/collector.go
+++ b/pkg/resourcesets/collector.go
@@ -105,7 +105,7 @@ func (h *ResourceHandler) GatherResources(ctx context.Context, resourceSelectors
 }
 
 func (h *ResourceHandler) gatherResourcesForGroupVersion(filter v1.ResourceSelector) ([]k8sv1.APIResource, error) {
-	var resourceList, resourceListFromRegex, resourceListFromNames []k8sv1.APIResource
+	var resourceList []k8sv1.APIResource
 
 	groupVersion := filter.APIVersion
 	logrus.Infof("Gathering resources for groupVersion: %v", groupVersion)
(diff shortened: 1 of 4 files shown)


pkg/resourcesets/collector.go
```diff
@@ -248,157 +174,147 @@ func (h *ResourceHandler) filterByLabel(ctx context.Context, dr dynamic.Resource
...
+			return filteredObjects, err
 		}
-
-		if len(filteredByNamespaceRegex) == 0 {
-			// none matched regex
-			// return whatever was filtered by exact namespace match
-			// if that list is also empty, it means no namespaces matched the given filters
-			return filteredByNamespace, nil
+	}
+	if filter.ExcludeResourceNameRegexp != "" {
```
mallardduck, line 292: This section of code is within `filterByNamespace` so we shouldn't be touching `ExcludeResourceNameRegexp` anymore. Since that's only necessary in the `filterByName` method. Think we need 292-299 removed.

pkg/resourcesets/collector.go
```diff
@@ -248,157 +174,147 @@ func (h *ResourceHandler) filterByLabel(ctx context.Context, dr dynamic.Resource
...
-		} else {
-			filteredByName = filteredByResourceNames
 		}
-
 	}
+
 	return filteredByName, nil
 }
 
 func (h *ResourceHandler) filterByNamespace(filter v1.ResourceSelector, filteredByName []unstructured.Unstructured) ([]unstructured.Unstructured, error) {
```
mallardduck, line 276: For this method we only concern with:

```
	Namespaces                []string              `json:"namespaces,omitempty"`
	NamespaceRegexp           string                `json:"namespaceRegexp,omitempty"`
```

And have no exclude mechanism. So essentially the logic for this method should just be:

`MatchNamespacesList || MatchNamespaceRegex`

pkg/resourcesets/collector.go
```diff
@@ -248,157 +174,147 @@ func (h *ResourceHandler) filterByLabel(ctx context.Context, dr dynamic.Resource
...
+		return apiResources, nil
+	}
+	if filter.KindsRegexp != "" {
+		var err error
+		kindRegexp, err = regexp.Compile(filter.KindsRegexp)
+		if err != nil {
+			return resourceList, err
+		}
+	}
+	allowedNames := make(map[string]bool)
```
mallardduck, line 191: nit: might be better to call these `allowedKinds` & `disallowedKinds`.

pkg/resourcesets/collector.go
```diff
@@ -248,157 +174,147 @@ func (h *ResourceHandler) filterByLabel(ctx context.Context, dr dynamic.Resource
...
+	var kindRegexp *regexp.Regexp
+	if filter.KindsRegexp == "" && len(filter.Kinds) == 0 {
+		// if no filters for resource kind are given, return entire resource list
+		return apiResources, nil
+	}
+	if filter.KindsRegexp != "" {
+		var err error
+		kindRegexp, err = regexp.Compile(filter.KindsRegexp)
+		if err != nil {
+			return resourceList, err
```
mallardduck, line 188: (unsure): In this scenario we're returning an empty slice and an `err`. I haven't tested this to confirm myself, but from an end-user POV we def want this type of error to cause the backup to fail completely. And ideally provide logging/error messages that will direct us to the (in this case) filter with Regex that caused errors.

In other words, if we have to chose between:
1. An error causing an empty backup to be taken and the backup marked completed, or
2. A loud noisy error and failed backup.
We certainly want to see option 2 be the outcome.

mallardduck: (this shouldn't be a blocker of this PR per say, just a general curiosity I have that I haven't tested in a lab to check the results yet)

pkg/resourcesets/collector.go
```diff
@@ -248,157 +174,147 @@ func (h *ResourceHandler) filterByLabel(ctx context.Context, dr dynamic.Resource
...
+	for _, name := range filter.ExcludeKinds {
+		disallowedNames[name] = true
+	}
+
+	// "resources" list has all resources under given groupVersion, first filter based on KindsRegexp
+	// Look for a match in either `Kind` (singular name) or `Name` (plural name).
+	// If an exclusion includes either of `Kind` or `Name`, then it's excluded, even if we matched on the other.
+	for _, resObj := range apiResources {
+		includeIt := allowedNames[resObj.Kind] || allowedNames[resObj.Name]
+		if !includeIt && kindRegexp != nil {
```
mallardduck, line 205: Not sure I'm following all the filtering logic here, it seems to be doing more than I would expect. Unless I'm wrong I think our psuedo-logic here is:

```
IncludeObject = (Match Kinds || !Match ExcludeKinds && Match KindsRegex)
```

Since I think it would be illogical to support users creating this:

```
  - apiVersion: apiextensions.k8s.io/v1
    kinds:
      - clusters
      - pods
    excludeKinds:
      - pods
```

But I think , this should be supported still:
```
  - apiVersion: apiextensions.k8s.io/v1
    kindsRegexp: "^c"
    kinds:
      - pods
    excludeKinds:
      - clusterroles
```

mallardduck (changes requested):
Things seem to work alright in lab testing, however there are still a few necessary changes per comments.

pkg/resourcesets/collector.go
ericpromislow: True. I mixed up excludeResourceNameRegexp with a non-existent `excludeNamespaceRegexp`

pkg/resourcesets/collector.go
ericpromislow: I kept this from the previous code, but I agree with you: if we're returning an non-nil error, the first return value should usually be nil (because the caller should generally not have to deal with a partially correct result from a function that errors out).

pkg/resourcesets/collector.go
ericpromislow: Done

pkg/resourcesets/collector.go
ericpromislow: Sure

mallardduck (approved):
LGTM!

ericpromislow:
Done in PR #415

PR #413, opened 2024-02-01
Fix Cross-Arch builds on arm64

mallardduck:
Currently on an arm64 based system the binary will be arm64.
Then when the package script is run, it copies that arm64 binary into the image.
This is as expected when building native arm64/amd64 binaries and images.

However, when trying to build cross compiled binaries and images locally it falls apart.
So using `USE_DOCKER_BUILDX=1` locally the container image will end up amd64, but the binary inside arm64.
This PR is trying to work out the most simple way to fix that local build issue without affecting the results of drone or release builds.

To do this we add a new `CROSS_ARCH` build flag meant for local dev needing this functionality.
So running either `CROSS_ARCH=true make`, or `CROSS_ARCH=true make {SCRIPT}` will adjust for that use-case.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -63,7 +63,10 @@ If you want to interact with the Minio deployed in the cluster, you can use the 
 
 ### Building on Different Architectures
 
-Currently we only support building this chart for amd64 nodes. The binary built by `make build` is set to build for amd64 and we have also included a flag for users running different architectures when trying to package the code into a docker image. Setting `USE_DOCKER_BUILDX=1` will force the package script to use docker buildx, setting the target platform to build for amd64. If you do not have `docker-buildx` installed please reference [this page](https://docs.docker.com/buildx/working-with-buildx/).
+Currently, we only support building this chart fully for `amd64` nodes. The binary and image building scripts will build "arch native" by default.
+So when building on `amd64` the resulting binary and image will both be `amd64` - same for `arm64`. This is how the drone builds operate.
+
+We have also included a flag for users running different architectures when trying to package the code into an `amd64` docker image. Building the binary with `CROSS_ARCH=true` set will cause it to build `amd64` and `arm64` binaries, and then setting `USE_DOCKER_BUILDX=1` will force the package script to use docker buildx, setting the container's target platform to build for `amd64`. If you do not have `docker-buildx` installed please reference [this page](https://docs.docker.com/buildx/working-with-buildx/).
 
 
 ### Developer Uninstallation
--- a/scripts/build
+++ b/scripts/build
@@ -13,17 +13,28 @@ LINKFLAGS="-X main.Version=$VERSION"
 LINKFLAGS="-X main.GitCommit=$COMMIT $LINKFLAGS"
 if [ "$CROSS_ARCH" != "true" ]; then
   CGO_ENABLED=0 go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o bin/backup-restore-operator
-  if [ "$CROSS" = "true" ] && [ "$ARCH" = "amd64" ]; then
+  if [ "$CROSS" = "true" ]; then
     GOOS=darwin go build -ldflags "$LINKFLAGS" -o bin/backup-restore-operator-darwin
     GOOS=windows go build -ldflags "$LINKFLAGS" -o bin/backup-restore-operator-windows
   fi
 else
-  GOARCH=amd64 CGO_ENABLED=0 go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o bin/backup-restore-operator
-  CGO_ENABLED=0 go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o "bin/backup-restore-operator$SUFFIX"
-  if [ "$CROSS" = "true" ]; then
-    GOARCH=amd64 GOOS=darwin go build -ldflags "$LINKFLAGS" -o bin/backup-restore-operator-darwin
-    GOOS=darwin go build -ldflags "$LINKFLAGS" -o "bin/backup-restore-operator-darwin$SUFFIX"
-    GOARCH=amd64 GOOS=windows go build -ldflags "$LINKFLAGS" -o bin/backup-restore-operator-windows
-    GOOS=windows go build -ldflags "$LINKFLAGS" -o "bin/backup-restore-operator-windows$SUFFIX"
+  if [ "$ARCH" = "amd64" ]; then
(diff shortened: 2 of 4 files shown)


mallardduck:
And to compare the `CROSS=true` option, here's what I get "before" this change:
```
→ file ./bin/*
./bin/backup-restore-operator: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, Go BuildID=2iR-Op9gWT0gxQesXbRX/uk4_3jTzXkuheOkUE_rc/qgYBRoNGOWtONkPiDNAe/r2vPEehr8G_AcAGWG3B7, stripped
```

And here's after the change:
```
→ file ./bin/*
./bin/backup-restore-operator:               ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=WvFXnyrglTkh5_SI4Pwl/I-SvrRfzqIUrThO1SAwg/jv17LME-ZL36nK-Z39My/qHPX8kbCUNS_Qx8qrkC2, stripped
./bin/backup-restore-operator-darwin:        Mach-O 64-bit executable x86_64
./bin/backup-restore-operator-darwin-arm64:  Mach-O 64-bit executable arm64
./bin/backup-restore-operator-windows:       PE32+ executable (console) x86-64 (stripped to external PDB), for MS Windows
./bin/backup-restore-operator-windows-arm64: PE32+ executable (console) Aarch64 (stripped to external PDB), for MS Windows
./bin/backup-restore-operator-arm64:         ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, Go BuildID=de_5kaI01cRc9uSeSbMT/4gEbEJF9sCkEPOcU4hMv/qgYBRoNGOWtONkPiDNAe/yTa_zy5SfM0JtGt31_h1, stripped
```

Note: edited PR to use $SUFFIX to be consistent with existing behaviour.

scripts/build
```diff
@@ -11,8 +11,30 @@ if [ "$(uname)" = "Linux" ]; then
...
-if [ "$CROSS" = "true" ] && [ "$ARCH" = "amd64" ]; then
+if [ "$CROSS_ARCH" != "true" ]; then
+  CGO_ENABLED=0 go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o bin/backup-restore-operator
+  if [ "$CROSS" = "true" ]; then
     GOOS=darwin go build -ldflags "$LINKFLAGS" -o bin/backup-restore-operator-darwin
     GOOS=windows go build -ldflags "$LINKFLAGS" -o bin/backup-restore-operator-windows
-fi
+  fi
+else
+  if [ "$ARCH" = "amd64" ]; then
```
mallardduck, line 21: This could be a lot smaller of a change if we only want to support "cross arch" builds from ARM64->AMD64. However it felt like we should make it symmetrical to allow going the other direction. The primary use case for me needing this change to begin with could easily be inverted.

Especially with our continued investment into "edge" we may likely see more arm64 support needed overtime. And that will increase the odds of a developer compiling on `amd64` with their lab/target cluster being `arm64`.

scripts/build
```diff
@@ -11,8 +11,30 @@ if [ "$(uname)" = "Linux" ]; then
...
+    CGO_ENABLED=0 go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o "bin/backup-restore-operator$SUFFIX"
+    GOARCH=amd64 CGO_ENABLED=0 go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o bin/backup-restore-operator
+    if [ "$CROSS" = "true" ]; then
+      GOOS=darwin go build -ldflags "$LINKFLAGS" -o "bin/backup-restore-operator-darwin$SUFFIX"
+      GOARCH=amd64 GOOS=darwin go build -ldflags "$LINKFLAGS" -o bin/backup-restore-operator-darwin
+      GOOS=windows go build -ldflags "$LINKFLAGS" -o "bin/backup-restore-operator-windows$SUFFIX"
+      GOARCH=amd64 GOOS=windows go build -ldflags "$LINKFLAGS" -o bin/backup-restore-operator-windows
+    fi
+  fi
+fi
```
ericpromislow, line 40: There's a lot of repetition in this code. I rewrote it with an "XARCH" variable (for the other architecture and came up with this):
```

CGO_ENABLED=0 go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o bin/backup-restore-operator

# Set CROSS to build for other platforms
if [ "$CROSS" = "true" ]; then
  for OS in darwin windows ; do
      GOOS=$OS go build -ldflags "$LINKFLAGS" -o bin/backup-restore-operator-$OS
  done
fi

# Set CROSS_ARCH to build for other architectures (one only currently)

if [ "$CROSS_ARCH" == "true" ]; then
  case "$ARCH" in
    amd64) XARCH=arm64 ;;
    arm64) XARCH=amd64 ;;
    *) echo "Unsupported ARCH of $ARCH" 1>&2 ; exit 1
  esac
  GOARCH=$XARCH CGO_ENABLED=0 go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o bin/backup-restore-operator-$XARCH
  if [ "$CROSS" = "true" ]; then
    for OS in darwin windows ; do
      GOARCH=$XARCH GOOS=$OS go build -ldflags "$LINKFLAGS" -o bin/backup-restore-operator-${OS}-$XARCH
    done
  fi
fi
```

ericpromislow (changes requested):
Just a suggestion to simplify the build-script logic

DEVELOPING.md
```diff
@@ -63,7 +63,10 @@ If you want to interact with the Minio deployed in the cluster, you can use the
 
 ### Building on Different Architectures
 
-Currently we only support building this chart for amd64 nodes. The binary built by `make build` is set to build for amd64 and we have also included a flag for users running different architectures when trying to package the code into a docker image. Setting `USE_DOCKER_BUILDX=1` will force the package script to use docker buildx, setting the target platform to build for amd64. If you do not have `docker-buildx` installed please reference [this page](https://docs.docker.com/buildx/working-with-buildx/).
+Currently, we only support building this chart fully for `amd64` nodes. The binary and image building scripts will build "arch native" by default.
+So when building on `amd64` the resulting binary and image will both be `amd64` - same for `arm64`. This is how the drone builds operate.
```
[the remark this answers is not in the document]
mallardduck: Yeah, in this case I added a line-break for human editor benefits. I guess I could be making an assumption on where this is rendered, but most Markdown renderers only split to new paragraph if it sees `(space)(space)(new-line)` preceding the new line.

I can make it a single line again tho for cleaner diffing.

ericpromislow (approved):
Looks good

PR #415, opened 2024-02-13
Bring in same changes as in v3 and v4

ericpromislow:
- Fix greediness problem in resourcesets
- Add unit tests


ericpromislow:
Better to wait for PR #420 to be merged into `release/v4.0` and then pull that change in here with rebases.

PR #416, opened 2024-02-13
Refactor to get rid of appendTempUID

### LINKED ISSUE #411: [CodeQuality] Fix filtering on object's Kind or Name

mallardduck:
Fixes #411

This PR gets rid of the need to use a UID (either temporarily assigned, or from unstructured UID) when filtering by name.
Ultimately the way this was achieved is by checking each item in the list only once - i.e. iterating over the items a single time and checking relevant conditions, rather than iterating over multiple changing lists/maps across the `filterByName` method.


pkg/resourcesets/collector.go
```diff
@@ -187,47 +187,35 @@ func (h *ResourceHandler) gatherResourcesForGroupVersion(filter v1.ResourceSelec
...
-	filteredByName, err := h.filterByName(filter, filteredByLabelItems)
+	filteredByName, err := h.filterByName(filter, &filteredByLabel.Items)
 	if err != nil {
-		return filteredObjects, err
+		return nil, err
 	}
 
 	if res.Namespaced {
 		if len(filter.Namespaces) > 0 || filter.NamespaceRegexp != "" {
 			filteredByNamespace, err = h.filterByNamespace(filter, filteredByName)
```
ericpromislow, line 208: Lines 208 - 2111 can just be
```
return h.filterByNamespace(filter, filteredByName)
```

pkg/resourcesets/collector.go
```diff
@@ -187,47 +187,35 @@ func (h *ResourceHandler) gatherResourcesForGroupVersion(filter v1.ResourceSelec
...
 	// however, in practice filtering by label happens at an API level when we paginate the resources (creating our initial list)
 	filteredByLabel, err := h.filterByLabel(ctx, dr, filter)
 	if err != nil {
-		return filteredObjects, err
+		return nil, err
 	}
 
-	filteredByLabelItems := h.appendTempUID(filteredByLabel.Items)
-	filteredByName, err := h.filterByName(filter, filteredByLabelItems)
+	filteredByName, err := h.filterByName(filter, &filteredByLabel.Items)
```
ericpromislow, line 201: I don't see why `filtererByLabel` needs to be passed by reference, as its contents are modified by the callee, but not the wrapper itself.

pkg/resourcesets/collector.go
```diff
@@ -248,97 +236,50 @@ func (h *ResourceHandler) filterByLabel(ctx context.Context, dr dynamic.Resource
...
-			for _, resObj := range filteredByResourceNames {
-				key := resObj.Object["filterUid"].(string)
-				if !filteredByNameMap[key] {
-					filteredByName = append(filteredByName, resObj)
+	for resObjID := range *resourceObjectsList {
+		// Check if the name is in the includeNames list
+		name := (*resourceObjectsList)[resObjID].GetName()
+		if includeNameMap[name] {
+			filteredByName = append(filteredByName, (*resourceObjectsList)[resObjID])
+			logrus.Debugf("Including [%s] because it mathed a name in ResourceNames", name)
```
ericpromislow, line 267: typo: should be `matched`

pkg/resourcesets/collector.go
```diff
@@ -248,97 +236,50 @@ func (h *ResourceHandler) filterByLabel(ctx context.Context, dr dynamic.Resource
...
+		} else {
+			if filter.ResourceNameRegexp != "" || filter.ExcludeResourceNameRegexp != "" {
+				// We expect this to be true, when set and match or when nil (true by default)
+				includeMatch := includeRegex == nil || includeRegex.MatchString(name)
+				// We expect this to be true, only when regex match (false by default)
+				excludeMatch := excludeRegex != nil && excludeRegex.MatchString(name)
```
ericpromislow, line 273: You can move the `excludeMatch` check inside
```
if includeMatch {
  excludeMatch := excludeRegex != nil && excludeRegex.MatchString(name)
  if !excludeMatch {
     filteredByName = append(...)
     logrs.Debugf("Including...")
 }
 ```
 
 This saves doing a pattern-match when it will never be used.

File: pkg/resourcesets/collector.go [hunk 5]
```diff
@@ -248,97 +236,50 @@ func (h *ResourceHandler) filterByLabel(ctx context.Context, dr dynamic.Resource
...
-					filteredByName = append(filteredByName, resObj)
+	for resObjID := range *resourceObjectsList {
+		// Check if the name is in the includeNames list
+		name := (*resourceObjectsList)[resObjID].GetName()
+		if includeNameMap[name] {
+			filteredByName = append(filteredByName, (*resourceObjectsList)[resObjID])
+			logrus.Debugf("Including [%s] because it mathed a name in ResourceNames", name)
+		} else {
+			if filter.ResourceNameRegexp != "" || filter.ExcludeResourceNameRegexp != "" {
+				// We expect this to be true, when set and match or when nil (true by default)
```
ericpromislow, line 270: I don't think you need either of these `We expect` comments, because they just echo what the code does, and the code's clear enough.

tests/stubs/README.md
```diff
@@ -3,4 +3,20 @@
 The files here are used for the unit tests for BRO code that's applicable. 
 Each file correlates to a unit test method and contains the mock data that will be used for filtering.
 Because each test method is targeting a specific filter scenario, we need the results to be uniform and assertable.
-This is the main reason we opt to create a new set of YAML for each test method and scenario.
\ No newline at end of file
+This is the main reason we opt to create a new set of YAML for each test method and scenario.
+
+## Unit Test Function & Stub File Example
+
```
ericpromislow, line 9: I think you can simplify this explanation (everything under "#Unit Test Function & Stub File Example), like so:

There's a 1:1 correspondence between each test function in `collector_test.go` and its input YAML in `test/stubs`. For example, the test `TestFilterByName_ExcludeRegexWithWildcardInclude` will get its YAML input from `test/stubs/TestFilterByName_ExcludeRegexWithWildcardInclude.yaml`, and that YAML is then unmarshalled into a slice of `unstructured.Unstructured` objects, which the filters expect to see. Each test constructs its own `v1.ResourceSelector` and verifies that the objects filtered by that selector are as expected.

ericpromislow (changes requested):
Mostly clarifications and typos in the comments, and a request for some simplification in the code.

tests/stubs/README.md
mallardduck: Updated the explanation to be a lot more in line with this one. LMK if you'd like it to be exactly word for word, or if you'd like other edits.

pkg/resourcesets/collector.go
```diff
@@ -249,96 +232,49 @@ func (h *ResourceHandler) filterByLabel(ctx context.Context, dr dynamic.Resource
...
-		}
-		if len(filteredByNameMap) > 0 {
-			// avoid duplicates
-			for _, resObj := range filteredByResourceNames {
-				key := resObj.Object["filterUid"].(string)
-				if !filteredByNameMap[key] {
-					filteredByName = append(filteredByName, resObj)
+	for resObjID := range resourceObjectsList {
+		// Check if the name is in the includeNames list
+		name := (resourceObjectsList)[resObjID].GetName()
```
ericpromislow, line 259: Last change: there's no need to parenthesize `resourceObjectsList`

pkg/resourcesets/collector.go
```diff
@@ -249,96 +232,49 @@ func (h *ResourceHandler) filterByLabel(ctx context.Context, dr dynamic.Resource
...
+		if includeNameMap[name] {
+			filteredByName = append(filteredByName, (resourceObjectsList)[resObjID])
```
ericpromislow, line 261: No parens needed

pkg/resourcesets/collector.go
```diff
@@ -249,96 +232,49 @@ func (h *ResourceHandler) filterByLabel(ctx context.Context, dr dynamic.Resource
...
+			logrus.Debugf("Including [%s] because it matched a name in ResourceNames", name)
+		} else {
+			if filter.ResourceNameRegexp != "" || filter.ExcludeResourceNameRegexp != "" {
+				includeMatch := includeRegex == nil || includeRegex.MatchString(name)
+
+				if includeMatch {
+					excludeMatch := excludeRegex != nil && excludeRegex.MatchString(name)
+					if !excludeMatch {
+						filteredByName = append(filteredByName, (resourceObjectsList)[resObjID])
```
ericpromislow, line 270: And no parens needed here

alexandreLamarre (commented):
Overall looks good to me. One nit for the current PR is to just rebase & squash the current PR commits into a single commit for readability

---- 

Comments about the overall test structure included in https://github.com/rancher/backup-restore-operator/pull/406  (no need to follow up on any of this in this PR),  just jotting down some thoughts:

I'd discourage the use of implicit details like matching test files to test names, which require people to read extra documentation to understand the tests/code at a glance. I only bring this up since this series of PRs are related to code quality improvements

One idiomatic go approach that comes to my mind is to combine the usage of [go:embed directives](https://pkg.go.dev/) in the test directory along the lines of : 


```go
package testdata

import (
	"embed"
	"io/fs"
	"path"
)

//go:embed stubs
var TestStubsFS embed.FS

func TestStubs(filename string) []byte {
	data, err := fs.ReadFile(TestStubFS, path.Join("stubs", filename))
	if err != nil {
		panic(err)
	}
	return data
}

// you could have another top level function taking the go:embed file and directly parsing
// it unstructured.Unstructured, etc...
```

This allows test data to be accessed from any path without having to rely on relative paths, which allows for more unit tests to consume test data transparently, along with a couple of other other minor benefits

One thing we would need to pay attention to with this particular solution is to add go build flags to the test file embedding the test data, to make sure its excluded from the binary being shipped

ericpromislow:
> One thing we would need to pay attention to with this particular solution is to add go build flags to the test file embedding the test data, to make sure its excluded from the binary being shipped

I thought `go build` skips any files matching `*_test.go`, and I don't see a reason to put `embed` directives in non-test source files.

alexandreLamarre:
> > One thing we would need to pay attention to with this particular solution is to add go build flags to the test file embedding the test data, to make sure its excluded from the binary being shipped
> 
> I thought `go build` skips any files matching `*_test.go` - I verified this by building and running `strings bin/backup-restore-operator | grep string-from-tests` and it didn't hit, while grepping for a string in non-test files hits.

sorry my train of thought was a little unclear, usually I stick those top level test functions in a regular go package because usually im building a non-test binary to run a test-environment, but here you're right a test package is more than sufficient

ericpromislow (changes requested):
Everything's good except `resourcesets/testdata/README.md` is now out of date. We prob don't need it at all.

ericpromislow (approved):
Looks good

pkg/resourcesets/collector.go
```diff
@@ -249,156 +229,79 @@ func (h *ResourceHandler) filterByLabel(ctx context.Context, dr dynamic.Resource
...
-				if !nameMatched {
-					logrus.Debugf("Skipping [%s] because it did not match ResourceNameRegexp [%s]", name, filter.ResourceNameRegexp)
-					continue
-				}
-			}
-			filteredByName = append(filteredByName, resObj)
-			key := resObj.Object["filterUid"].(string)
-			filteredByNameMap[key] = true
-		}
+		includeRegex = regexp.MustCompile(filter.ResourceNameRegexp)
```
alexandreLamarre, line 248: I don't know if we have a pre-processing pass elsewhere in the code that validates the user input filters before it reaches this function, otherwise we probably want to avoid panicking if the regex is invalid.

pkg/resourcesets/collector.go
```diff
@@ -249,156 +229,79 @@ func (h *ResourceHandler) filterByLabel(ctx context.Context, dr dynamic.Resource
...
-			allowedNames[name] = true
-		}
-		for _, resObj := range resourceObjectsList {
-			metadata := resObj.Object["metadata"].(map[string]interface{})
-			name := metadata["name"].(string)
-			if allowedNames[name] {
-				filteredByResourceNames = append(filteredByResourceNames, resObj)
-			}
-		}
+		excludeRegex = regexp.MustCompile(filter.ExcludeResourceNameRegexp)
```
alexandreLamarre, line 252: same consideration as above

ericpromislow: We don't have a preprocess pass to verify the regexes, but they're embedded in the chart, so it's unlikely that users will change or add to them. However this should use `regex.Compile` and return a non-nil error object through the normal channels so it can be surfaced gracefully (and if there is an invalid regex, it should be caught during integration testing, because running a backup will try to compile every submitted regex.

alexandreLamarre: ok cool

alexandreLamarre (approved):

PR #419, opened 2024-02-23
Rebase resource-set refactoring from 3.0 to 4.0

ericpromislow:
And ensure this chart works with rancher 2.8


ericpromislow:
Too many changes here outside the resourceset part don't make sense.

PR #421, opened 2024-02-28
Move cross arch fix to v4

ericpromislow:
This is working on release/v3.0, needs to get ported to 4.0 (and then 5.0)


ericpromislow:
The diff makes no sense.

PR #426, opened 2024-03-21
[v5.x] Update renovate to use correct branch

mallardduck:


ericpromislow (commented):
LGTM - If we're running this on `release/v3.0` it should be on the other live branches as well.

PR #427, opened 2024-03-21
[release/v3.0] Fix the branches on renovate config

mallardduck:
After reading docs for renovate config I realized the config should be uniform across active branches. So I'm adding all 3 versions to the v3 branch config first. Then these can be cherry-picked/merged up to v4 and v5 as desired.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -3,7 +3,9 @@
     "github>rancher/renovate-config#release"
   ],
   "baseBranches": [
-    "release/v3.0"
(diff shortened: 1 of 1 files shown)


mallardduck:
Maybe I need to point this at the release/v5.0 branch though - since GH actions source via default branch. From the looks of it, that would make sense from the public perspective on the repo. I don't have access to see more as I'm still waiting on team adjustments for my account.

ericpromislow (approved):
LGTM

PR #430, opened 2024-04-01
Add upstream-version to release workflows

mallardduck:
This is a small improvement I think should be in all the active branches. And while in the future we'll probably work from newer-backwards, right now we're still in a place where starting here makes sense to me.

Since I have plans to create subsequent syncing PRs up from 3->4->5 in order to properly catch all the changes 3.x branch has currently. This should both help address the renovate situation and to address #411


ericpromislow (approved):
Looks good, local tests pass

PR #429, opened 2024-03-28
Update minio-go and wrangler libraries

mallardduck:
Also bump golang per requirements of other updates.

Tested on Cluster with:

| Runtime       | Node Count     | Rancher Version | BRO Chart |
| ------------- | ------------ | --------------- | ------------------------------ |
| v1.26.11+k3s1 | 6 (3workers) | v2.7.11         | rancher-backup:102.0.3+up3.1.3 |

### Test Steps

- Took a backup (B1)
- Swapped out backups image for my PR image,
- Took a new backup (B2),
- Changed local rancher cluster name,
- Took another backup (B3)
- Restored to backup taken before using new image (B1),
- Verify changes restored as expected,
- Restore backup taken after name change (B3) to go back to "newest",
- Verify changes,
- restore backup taken on new image before name change (B2)

---

### Rancher deps (wrangler and lasso)

Per our readme, current BRO releases should only need to target Rancher 2.7.x and up.
> the tag v3.x.x is cut from the release/v3.0 branch for Rancher v2.7.x line

Looking at `rancher/wrangler` it seems we are safe to (and should) bump Wrangler to match this minimum. Per Wrangler docs, 1.0.x was meant for 2.6.x and 1.1.x for 2.7.x. Similarly lasso looks to be pretty safe as all those changes are rather minor.

### K8s.io deps

For all the k8s.io dependencies that updated we should be pretty good for compatibility. Since it's all client code tied to specific k8s server versions - the biggest risk is lacking feature support for k8s newer than the one we use. And potentially more concerning removal of features from the client that old k8s may still have.

PR #431, opened 2024-04-01
Merge release/v3.0 into release/v4.0

mallardduck:
This should be on hold and then rebased to merge after:

- #429 
- #430


mallardduck:
🤔 - I intentionally did this as a squashed commit locally, wasn't expecting this many commits when I pushed and created PR. Either way as mentioned in description this should be on hold until those others merge then i'll rebase at that point in time and address this oddness.

ericpromislow (approved):
LGTM

PR #432, opened 2024-04-01
Merge release/v4.0 into release/v5.0

mallardduck:
This should be on hold and then rebased to merge after:

- #431
- #429 
- #430


ericpromislow (approved):
We've seen and tested all these changes in 3.0, and I just verified they work putting everything together into v5.0

PR #439, opened 2024-04-05
Update dependencies for 3.x branch

mallardduck:
So far this attempts to roll all into one:

- https://github.com/rancher/backup-restore-operator/pull/433 (a less breaking version of)
- https://github.com/rancher/backup-restore-operator/pull/442
- https://github.com/rancher/backup-restore-operator/pull/440
- https://github.com/rancher/backup-restore-operator/pull/438
- https://github.com/rancher/backup-restore-operator/pull/441 (except for tests which need to be lower version to match hull)
- https://github.com/rancher/backup-restore-operator/pull/437
- https://github.com/rancher/backup-restore-operator/pull/443
- https://github.com/rancher/backup-restore-operator/pull/435 (a less breaking version of)
- https://github.com/rancher/backup-restore-operator/pull/436
- https://github.com/rancher/backup-restore-operator/pull/445
- https://github.com/rancher/backup-restore-operator/pull/444
- https://github.com/rancher/backup-restore-operator/pull/446
- https://github.com/rancher/backup-restore-operator/pull/447 (a less breaking version of)

## Concerns for 3.x Branch (Supporting Rancher 2.7.x)
Essentially every instance where a "less breaking version" is noted, it is applied to prevent `k8s.io` from bumping to 28.x which fully removes PSPs. And in turn would break the minimum k8s version supported by rancher.

### What was fixed, or what change have occurred
A lot of dependency bumps.

### Areas or cases that should be tested

- Rancher backups on clusters with PSPs enabled still.
- Backups (and restores) with encryption enabled.

I have done a good amount of testing on clusters without PSP enabled - however am not equipped to test with PSPs enabled so this is a blind spot. Without PSPs though I have tested both the local backup and restore, and the "migration restore" based on a backup from a stable BRO image.

### What areas could experience regressions

- Encryption related code paths; because the crypto lib was bumped
- PSP related code paths; because the k8s.io/* packages were bumped (though just shy of removing PSP code)

### Are the repro steps accurate/minimal?
Yes. I've done minimal testing using the following testing methods (however my tests were without PSP and without encryption).

#### Local Backup/Restore
This should be done two times; once with PSPs enabled and once with encryption enabled (2 different test runs).

1. Setup k8s cluster and rancher per requirements,
2. Install backups,
3. Create a one-off backup,
4. Change the name of the local rancher cluster (or some other change),
5. Run the restore,
6. Observe your change will have been un-done.

#### Migration Restore
This should be done two times; once with PSPs enabled and once with encryption enabled (2 different test runs).

1. Setup k8s cluster and rancher per requirements,
2. Install backups,
3. Change the name of the local rancher cluster (or something else, in my case I create a downstream),
4. Create a one-off backup pushing to S3,
5. Nuke the nodes the cluster is on,
6. Reinstall k8s cluster (but not rancher),
7. Run migration restore per docs,
8. Install cert-manager and Rancher

## Next Steps

Once tested and merged these changes should be "rolled up" to 4.x and 5.x. And then subsequently 4.x can under-go even more version bumps without concern for dropping PSP support.


ericpromislow (approved):
Looks fine, my unit and integration tests pass

MKlimuszka:
These are CVE bumps. Will be covered by general chart testing.

PR #455, opened 2024-04-11
Migration to GHA

alexandreLamarre:
- [x] Adds goreleaser for binary release CI, a standard release tool for go binaries. Preserves the same static link flags as our `package` script
- [x] Migrates main CI (build & test) to gha dynamic matrices (linuxamd, linuxarm) x (min_supported_k8s-version, max_supported_k8s_version)
- [x] Migrates image building CI to using docker-build-and-push action using buildx and docker manifests.

⚠️ Does not include arm64 support, will follow up in a separate PR on this

- [example goreleaser CI](https://github.com/alexandreLamarre/backup-restore-operator/actions/runs/8802582733)
- [example publish CI](https://github.com/alexandreLamarre/backup-restore-operator/actions/runs/8801741708/job/24156312703)

image
image


.goreleaser.yaml
```diff
@@ -0,0 +1,46 @@
...
+        - -s
+        - -X main.Version={{.Version}} -X main.GitCommit={{.Commit}}
+      flags:
+        - -trimpath
+      env:
+        - CGO_ENABLED=0
+archives:
+    - id: backup-restore-operator
+      builds:
+        - pushprox-client
```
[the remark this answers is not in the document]
alexandreLamarre: looks like I made a mistake in this config, when I was copying it over from PushProx

.github/workflows/scripts/setup-cluster.sh
```diff
@@ -0,0 +1,55 @@
...
+
+if [ -z "$K3S_VERSION" ]; then
+  echo "K3S_VERSION must be specified when setting up a cluster, use `k3d version list k3s` to find valid versions"
+  exit 1
+fi
+
+# waits until all nodes are ready
+wait_for_nodes(){
+  echo "wait until all agents are ready"
+  while :
```
ericpromislow, line 20: My preference is to support a timeout and fail if it's reached. I would go with 5 x an observed time on a non-state-of-the-art system.

scripts/integration
```diff
@@ -4,30 +4,10 @@ function echo_with_time {
...
-
-k3s kubectl get nodes -o wide --show-labels
-
-docker image save rancher/backup-restore-operator:$TAG -o /tmp/bro.img
-
-k3s ctr images import /tmp/bro.img
-
 ls -la ./dist/artifacts
 
 # In case short commit only conists of numbers, it is regarded valid by Helm when packaging
```
ericpromislow, line 13: typo (I know not part of this change): `consists`

.github/workflows/scripts/setup-cluster.sh
```diff
@@ -0,0 +1,55 @@
...
+
+source ./scripts/version
+
+if [ -z "$CLUSTER_NAME" ]; then
+    echo "CLUSTER_NAME must be specified when setting up a cluster"
+    exit 1
+fi
+
+if [ -z "$K3S_VERSION" ]; then
+  echo "K3S_VERSION must be specified when setting up a cluster, use `k3d version list k3s` to find valid versions"
```
ericpromislow, line 13: Please use `$(k3d version list k3s)` instead of backquotes

scripts/integration
```diff
@@ -4,30 +4,10 @@ function echo_with_time {
     echo "$(date --utc +%Y-%m-%dT%H:%M:%SZ) "$@""
```
ericpromislow, line 4: Again I know it's not part of this change, but I would change `"$@"` to `$*` because it's just for an `echo` statement. And `$@` outside double-quotes (which is what's going here) is the same as `$*`.

ericpromislow (changes requested):
Just nits. The CI files look good, and it's helpful that the github actions are passing.

.github/workflows/scripts/setup-cluster.sh
```diff
@@ -0,0 +1,55 @@
...
+    # all nodes are ready; exit
+    if [[ $readyNodes == 1 ]]
+    then
+      break
+    fi
+    sleep 1
+  done
+}
+
+k3d cluster delete ${CLUSTER_NAME} || true
```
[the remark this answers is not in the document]
mallardduck: second adding `shellcheck` - i already have it in my IDE and generally try to follow its guidance (as a sidenote, I used it as I rewrote kubectl CI scripts).

mallardduck (approved):
LGTM.

@ericpromislow - IIRC you said you would handle this round of release stuff for BRO. So if we merge this before that release then this will most directly affect you, so i'll mostly defer to your call - beyond this initial review.

ericpromislow (approved):
Looks good now

PR #464, opened 2024-05-07
Adjust Renovate Rules

mallardduck:
This PR adds a few Rancher modules as global ignores across all branches. This will hopefully prevent the renovate PRs from causing as many breaking changes. Often when those modules are updated they need code generation to run too which is not currently part of Renovate tasks.

~This PR attempts to address the way that the 3.x and 4.x branch must keep K8S support to match their rancher versions. Essentially when the auto bump affects these packages (which are not versioned) the result is bumping peers (`k8s.io` items) that these branches of BRO cannot increase without causing breaking changes.~

Example Open PRs Causing issues:
https://github.com/rancher/backup-restore-operator/pull/461
https://github.com/rancher/backup-restore-operator/pull/462

And here's an example of what it takes to get these types of PRs mergable:
https://github.com/rancher/backup-restore-operator/pull/458
https://github.com/rancher/backup-restore-operator/pull/459

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -3,6 +3,20 @@
     "github>rancher/renovate-config#release",
     "group:all"
   ],
+  "packageRules": [
+    {
+      "packageNames": [
+        "github.com/rancher/hull",
+        "github.com/rancher/lasso",
+        "github.com/rancher/wrangler/v2"
+      ],
+      "enabled": false,
+      "branches": [
+        "release/v3.0",
+        "release/v4.0"
+      ]
+    }
+  ],
   "baseBranches": [
     "release/v3.0",
     "release/v4.0",
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -11,6 +11,8 @@
   "ignoreDeps": [
     "github.com/rancher/hull",
     "github.com/rancher/lasso",
-    "github.com/rancher/wrangler/v2"
+    "github.com/rancher/wrangler",
+    "github.com/rancher/wrangler/v2",
+    "github.com/rancher/wrangler/v3"
   ]
 }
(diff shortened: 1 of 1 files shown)


alexandreLamarre (changes requested):
Looks like a better way to do this would be to use https://docs.renovatebot.com/configuration-options/#ignoredeps. 

I think potentially we would want to ignore bumps to those in any branch of backup-restore-operator, since it likely requires manual work or aggressive updates that force support out of a given k8s range

An example from another big open source repo is:

https://github.com/open-telemetry/opentelemetry-collector/blob/c6b70a7ec794d1a296f82863032def0848f870e0/renovate.json#L69C1-L71C7

mallardduck:
@alexandreLamarre - thanks for the input here, that does seem like it will make more sense! Updated PR to use that.

.github/renovate.json
```diff
@@ -3,23 +3,14 @@
...
-  ],
   "baseBranches": [
     "release/v3.0",
     "release/v4.0",
     "release/v5.0"
+  ],
+  "ignoreDeps": [
+    "github.com/rancher/hull",
+    "github.com/rancher/lasso",
+    "github.com/rancher/wrangler/v2"
```
alexandreLamarre, line 14: Not sure if `github.com/rancher/wrangler` is in use in the older release branches, but if it is we should ignore that as well.

Also adding `github.com/rancher/wrangler/v3` ahead of time is not a bad idea here, since that is released or will release soon (i forget)

mallardduck: That's a great point I overlooked and some very solid foresight.

mallardduck: Wrangler 0-3 now included.

alexandreLamarre (approved):

PR #466, opened 2024-05-09
[dev-2.9][BRO release/v5.0] Bump k8s to 1.29.3

ericpromislow:
Related to: [#44097](https://github.com/rancher/rancher/issues/44097)

This doesn't completely bump the BRO controller to k8s 1.29.3 because the code
uses a dropped exported function from k8s.io/apiserver to get the encryption transformer.

Quoting the k8s.io/apiserver readme:

> ~~## Things you should *NOT* do
> 
>  1. Directly modify any files under `pkg` in this repo.  Those are driven from `k8s.io/kubernetes/staging/src/k8s.io/apiserver`.
>  2. Expect compatibility.  This repo is changing quickly in direct support of
>     Kubernetes and the API isn't yet stable enough for API guarantees.~~

This caveat is also subject to change.


mallardduck:
@ericpromislow - I think I never fully forward ported things from 3/4 up to 5. And that might account for fixing some of the stuff you're encountering here. You mention the removed method related to encryption, I assume this one:

https://github.com/rancher/backup-restore-operator/blame/04aa3e529489f1046443b5afe03d4abcadaebb53/pkg/util/util.go#L44

However on the 3/4 branch I've actaully already updated us to be using that code:
https://github.com/rancher/backup-restore-operator/blame/530cb6353dddf5b61879e37262f34023354e611a/pkg/util/util.go#L45

ericpromislow:
This PR is out of date, easier to create a new PR

PR #467, opened 2024-05-09
Bump to k8s 1.29.

ericpromislow:
The encryption getter now wants an APIServerID, but this field is currently used only for metrics, so we can pass an empty string.


pkg/util/util.go
```diff
@@ -25,24 +25,27 @@ const (
...
 	}
 	encryptionConfigBytes, ok := encryptionConfigSecret.Data[encryptionProviderConfigKey]
 	if !ok {
-		return transformerMap, fmt.Errorf("no encryptionConfig provided")
+		return nil, fmt.Errorf("no encryptionConfig provided")
 	}
 	err = ioutil.WriteFile(encryptionProviderConfigKey, encryptionConfigBytes, os.ModePerm)
 	defer os.Remove(encryptionProviderConfigKey)
+
+	// ctx context.Context, filepath string, reload bool, apiServerID string
```
mallardduck, line 42: Probably intended to be a few lines lower? Or just a temporary dev note?

mallardduck (commented):
Just one note.

pkg/util/util.go
ericpromislow: fixed

mallardduck (approved):

PR #470, opened 2024-05-13
Set default ci in master branch

alexandreLamarre:
So that newer long-lived branches can default to configurations in there. 

Remove default drone CI as well.


alexandreLamarre (commented):
The code in master is a little outdated, so I decided it may not be worth the trouble to keep it up to date with our active release branches, rather it is more important that the github CI files are present there to set sane defaults for future long-lived branches (or feature branches)

mallardduck:
I wonder if we just delete the `master` branch? Given that we develop on release branches, does it even make sense to use it at all? Or should we just plan to use w/e the highest release branch is to create the base for the next release branch? (I.e. `release/v5.0` will become the base for `release/v6.0` once we start on Rancher 2.10.x?)

Or are you thinking we use the master branch as an intermediate between releases? If so, how will we keep it in sync with the latest release branch?

ericpromislow:
I assumed once we start working on rancher 2.10 (or 3.0, or whatever comes after 2.9) we'll branch `release/v6.0` off of `release/v5.0` and continue development there. And we can leave `master` for anything that wants a default branch that isn't tied to a release.

PR #457, opened 2024-05-01
fix typo in local ref for CI actions

alexandreLamarre:
- embedded job did not point to the correct path
- embedded job was not callable without `workflow_call`

PR #471, opened 2024-05-13
[Backport][v2.8] GHA CI

alexandreLamarre:
Backport of :
https://github.com/rancher/backup-restore-operator/pull/455
https://github.com/rancher/backup-restore-operator/pull/457
https://github.com/rancher/backup-restore-operator/pull/469


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - "*"
+
+env:
+  REGISTRY: docker.io
+  REPO : rancher
+
+
+jobs:
+  ci :
+    uses: rancher/backup-restore-operator/.github/workflows/ci.yaml@release/v5.0
```
alexandreLamarre, line 15: Decided to leave this as reference to the current default branch, but open to changing it to reference the target branch `release/v4.0` -- applies to a couple of other places.

Applies to a couple of other places. Perhaps discussing/documenting backport rules for CI is a good idea to be addressed in a follow up PR?

ericpromislow: I'm ok with this. If some future version is no longer compatible with the version in `release/v5.0` we can redefine it in that branch, and then have future branches reference that one.
For example, say we have to redefine the CI body in `release/v6.0`. Then the line in `release/7.0` etc. would be `uses: rancher/backup-restore-operator/.github/workflows/ci.yaml@release/v6.0`

So then if we want to be consistent, maybe this CI body should be defined in `release/v3.0' and referenced in `v4` and `v5`? WDYT?

alexandreLamarre: Yeah that sounds like a good idea, `release/v3.0` is probably the best bet, but we could also consider referencing `master`

Another idea is to have an orphan branch that specifically holds the CI files... Although that could complicate things more than help, but we could design around that to make it easier to work with

.github/workflows/scripts/supported-versions.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+# Prints supported versions based on the current release branch targeted
+# Version output is in JSON
+
+set -e
+set -x
+
+if git merge-base --is-ancestor origin/release/v5.0 HEAD
+then
+  echo -n "[\"v1.23.9-k3s1\", \"v1.29.3-k3s1\"]"
```
ericpromislow: If it's available, I would prefer
```
printf '["%s", "%s"]' v1.23.9-k3s1 1.29.3-k3s1
```
etc, for the other two lines. No backslashes, although same number of characters.

alexandreLamarre: Sure np, this applies to the other branches as well, so I can make that change to all 3 in  https://github.com/rancher/backup-restore-operator/pull/473, then cherry-pick that commit to the other backports

.goreleaser.yaml
```diff
@@ -0,0 +1,46 @@
+# This is an example .goreleaser.yml file with some sensible defaults.
```
ericpromislow, line 1: We might as well remove the generated comments. Since this file was generated, maybe replace the two lines with
```
# Generated by running `goreleaser` and then modified.
```

.goreleaser.yaml
```diff
@@ -0,0 +1,46 @@
...
+# Make sure to check the documentation at https://goreleaser.com
+before:
+  hooks:
+    # You may remove this if you don't use go modules.
```
ericpromislow, line 5: Might as well remove this comment too.

.goreleaser.yaml
alexandreLamarre: Sure, I can strip all non-relevant comments and link to the upstream documentation

.goreleaser.yaml
```diff
@@ -0,0 +1,46 @@
...
+snapshot:
+  name_template: "{{ incpatch .Version }}-next"
+changelog:
+  sort: asc
+  filters:
+    exclude:
+      - '^docs:'
+      - '^test:'
+
+# The lines beneath this are called `modelines`. See `:help modeline`
```
ericpromislow, line 43: The two explanatory comments can also be removed.

ericpromislow (changes requested):
Just some minor suggestions -- I tested the changes I use locally (like `scripts/deploy` and they're all good

ericpromislow (dismissed):
Will need same changes happening on release/v5.0 re the published binaries and charts tgz

.github/workflows/scripts/supported-versions.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+# Prints supported versions based on the current release branch targeted
+# Version output is in JSON
+
+set -e
+set -x
+
+if git merge-base --is-ancestor origin/release/v5.0 HEAD
+then
+  echo -n "[\"v1.23.9-k3s1\", \"v1.29.4-k3s1\"]"
```
mallardduck, line 10: This is slightly different than the 2.9 version - since this script seems to cover all releases, should we make sure the script is the same on each branch?

mallardduck (approved):

PR #472, opened 2024-05-13
[Backport][v2.7] GHA CI

alexandreLamarre:
Backport of :
https://github.com/rancher/backup-restore-operator/pull/455
https://github.com/rancher/backup-restore-operator/pull/457
https://github.com/rancher/backup-restore-operator/pull/469


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - "*"
+
+env:
+  REGISTRY: docker.io
+  REPO : rancher
+
+
+jobs:
+  ci :
+    uses: rancher/backup-restore-operator/.github/workflows/ci.yaml@release/v5.0
```
alexandreLamarre, line 15: Same comment as the 2.8 backport
> Decided to leave this as reference to the current default branch, but open to changing it to reference the target branch release/v4.0 -- applies to a couple of other places.
> 
> Applies to a couple of other places. Perhaps discussing/documenting backport standards for CI is a good idea to be addressed in a follow up PR?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,88 @@
+name: Backup Restore CI
+
+permissions:
+  contents : read
+
+on:
+  workflow_call:
+  push:
+    branches:
+      - release/v5.0
```
ericpromislow, line 10: Shouldn't this just be `release/v*` so we don't have to keep updating it?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,88 @@
...
+    - 'docs/**'
+    - '*.md'
+    - '.gitignore'
+    - 'CODEOWNERS'
+    - 'LICENSE'
+jobs:
+  build:
+    strategy:
+      matrix:
+        os : [
```
ericpromislow, line 30: extra white-space at the end

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,88 @@
...
+    paths-ignore:
+    - 'docs/**'
+    - '*.md'
+    - '.gitignore'
+    - 'CODEOWNERS'
+    - 'LICENSE'
+jobs:
+  build:
+    strategy:
+      matrix:
```
ericpromislow, line 29: Alternative to run on both platforms:
```
 strategy:
      matrix:
        archBox:
        - { arch: amd64, vmArch: x64 }
        - { arch: arm64, vmArch: arm64 }
    runs-on: runs-on,runner=1cpu-linux-${{ matrix.archBox.vmArch }},run-id=${{ github.run_id }}
    ```

File: .github/workflows/ci.yaml [hunk 5]
```diff
@@ -0,0 +1,88 @@
...
+        K3S_VERSION : ${{ fromJSON(needs.build.outputs.k3sversions) }}
+    steps:
+      - name : Checkout repository
+        uses : actions/checkout@v4
+      - name : Fetch build artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: bro-build-artifacts-${{ matrix.os }}
+          path: ./dist/artifacts/
+      - name : Debug
```
ericpromislow, line 75: I use this too, but I assume we want to pull out the Debug parts once everything's working

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - name : "Read vault secrets"
+        uses : rancher-eio/read-vault-secrets@main
+        with:
+          secrets: |
+            secret/data/github/repo/${{ github.repository }}/dockerhub/rancher/credentials username | DOCKER_USERNAME ;
+            secret/data/github/repo/${{ github.repository }}/dockerhub/rancher/credentials password | DOCKER_PASSWORD
+      - name : Checkout repository
+        uses: actions/checkout@v4
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
```
ericpromislow, line 37: Extra white-space at end of line 37

.github/workflows/scripts/setup-cluster.sh
```diff
@@ -0,0 +1,70 @@
...
+    if [[ $readyNodes == 1 ]]
+    then
+      break
+    fi
+    sleep 1
+  done
+}
+
+k3d cluster delete $CLUSTER_NAME || true
+k3d cluster create $CLUSTER_NAME --image "docker.io/rancher/k3s:${K3S_VERSION}"
```
ericpromislow, line 58: The `docker.io/` part isn't needed

.github/workflows/scripts/setup-cluster.sh
```diff
@@ -0,0 +1,70 @@
...
+  echo "wait until all agents are ready"
+  while :
+  do
+    current_time=$(date +%s)
+    elapsed_time=$((current_time - start_time))
+    if [ $elapsed_time -ge $timeout ]; then
+        echo "Timeout reached, exiting..."
+        exit 1
+    fi
+
```
ericpromislow, line 30: Space on line 30 not needed (`git diff` is complaining about these)

package/Dockerfile
```diff
@@ -1,3 +1,10 @@
...
+WORKDIR /usr/src/app
+COPY go.mod go.sum ./
+RUN go mod download
+COPY . .
+RUN ./scripts/build
+
 FROM registry.suse.com/bci/bci-micro:latest
-COPY bin/backup-restore-operator /usr/bin/
-ENTRYPOINT ["backup-restore-operator"]
+COPY --from=builder /usr/src/app/bin/backup-restore-operator  /usr/bin/
```
ericpromislow, line 9: extra space after `backup-restore-operator`

scripts/integration
```diff
@@ -1,36 +1,16 @@
 #!/bin/bash
 set -ex
 function echo_with_time {
-    echo "$(date --utc +%Y-%m-%dT%H:%M:%SZ) "$@""
+    echo "$(date --utc +%Y-%m-%dT%H:%M:%SZ) $*"
 }
 
-source $(dirname $0)/version
+source ./scripts/version
```
ericpromislow, line 7: I thought leaving it as `$(dirname $0)` would be better in case you're running this from some other directory

scripts/integration
```diff
@@ -1,36 +1,16 @@
 #!/bin/bash
 set -ex
 function echo_with_time {
-    echo "$(date --utc +%Y-%m-%dT%H:%M:%SZ) "$@""
+    echo "$(date --utc +%Y-%m-%dT%H:%M:%SZ) $*"
```
ericpromislow, line 4: The `--utc` option isn't recognized on Darwin (`/bin/date` so I'm not running some variant) but `-u` is a shortcut for it on linux. So better to just go with `-u`

ericpromislow (changes requested):
Just white-space issues, although you might want to try the dual-runner solution and see if CI succeeds on both platforms.

mallardduck:
I think that technically for both of these backports, we don't even need to bring the `.github` folder over. GHA always finds workflows based on the current default branch. And we're not likely to ever "rollback" and use an older branch as default, so they'll just be unused blobs of data on old branches.

As we move forward the new base branch (eventually `release/v6.0`) will inherit the same workflows we have when we branch. And for example, if we need different release processes for 6.0 while still using "old" workflows for old branches we would maintain both workflows in parallel and differ based on branches/tags that trigger them.

What is probably worth backporting though are mainly the parts that remove drone and bring the new build scripts. Since I think only scripts we put under `.github` will be in the context of the running action by default. And other scripts (like `./scripts` folder) are accessed via what the GHA clones.

mallardduck (approved):

Issue #474, opened 2024-05-13
GHA doesn't create release artifacts that the charts expect


ericpromislow:
This is on BRO `release/v5 v5.0.0-rc5` with the GHA changes.

`make prepare VERSION=$VERSION PACKAGE=rancher-backup` fails with this error message:

```
INFO[0000] Pulling https://github.com/rancher/backup-restore-operator/releases/download/v5.0.0-rc5/rancher-backup-5.0.0-rc5.tgz from upstream into charts
FATA[0000] encountered error while preparing main chart: encountered error while trying to pull upstream into charts: unable to read gzip formatted file: gzip: invalid header
```

Here are the assets for this release:
```
https://github.com/rancher/backup-restore-operator/releases/download/v5.0.0-rc5/backup-restore-operator_5.0.0-rc5_darwin_amd64.tar.gz
…/backup-restore-operator_5.0.0-rc5_linux_amd64.tar.gz 
…/backup-restore-operator_5.0.0-rc5_windows_amd64.tar.gz 
…/backup-restore-operator_5.0.0-rc5_darwin_arm64.tar.gz 
…/backup-restore-operator_5.0.0-rc5_linux_arm64.tar.gz 
…/backup-restore-operator_5.0.0-rc5_windows_arm64.tar.gz 
```

Here are the assets for the release from [v4.0.2-rc1](https://github.com/rancher/backup-restore-operator/releases/tag/v4.0.2-rc1):
```
https://github.com/rancher/backup-restore-operator/releases/download/v4.0.2-rc1/rancher-backup-4.0.2-rc1.tgz
https://github.com/rancher/backup-restore-operator/releases/download/v4.0.2-rc1/rancher-backup-crd-4.0.2-rc1.tgz
https://github.com/rancher/backup-restore-operator/releases/download/v4.0.2-rc1/backup-restore-operator-amd64
https://github.com/rancher/backup-restore-operator/releases/download/v4.0.2-rc1/backup-restore-operator-arm64
```

The 4.0.2 tgz files are gzip-tarred charts, and the binaries are for linux. I guess the binary part can be fixed in the `.goreleaser.yaml` file, but we need a github action that tgz's the chart subdirectories.

PR #475, opened 2024-05-13
fix generated binary artifact

ericpromislow:
# DO NOT MERGE

## Just wanted to test a change and see what happens in GHA

Related to: https://github.com/rancher/backup-restore-operator/issues/474


mallardduck:
I think we don't actually need to produce binary files for the charts. Since the chart should pull in the image and I think @alexandreLamarre has the images properly producing arm64 and amd64 builds.

However, if we do want to export binary files and archives (both) we can simply add a new entity under `archives`. And once you have more than 1 entity we need to add an `id`.

Something like:
```yaml
  - id: bin-build
    format: binary
    name_template: >-
      {{ .Binary }}_
      {{- tolower .Os }}_{{ .Arch }}
      {{- if .Amd64 }}-{{ .Amd64 }}{{ end }}
      {{- if .Arm }}v{{ .Arm }}{{ end }}{{ end }}
```

I think the main issue with charts is that it's expecting a tar.gz for backups still. So if we update `package.yaml` on the charts repo we might not need to change anything here.

Specifically I think if we change the file to look more like:
```yaml
url: https://github.com/rancher/backup-restore-operator.git
subdirectory: charts/rancher-backup
commit: {commit has matching release version}
version: 104.0.0
```

And then do the same for `rancher-backup-crd` chart. After those changes I think we can get charts build working correctly again without actually packaging and "releasing" the chart with each release.

However I'll defer to @alexandreLamarre since he did the GHA migration work for BRO repo. So he can confirm/deny these as good or bad ideas.

alexandreLamarre:
Superseded by https://github.com/rancher/backup-restore-operator/pull/476

ericpromislow:
See https://github.com/rancher/backup-restore-operator/pull/475#issuecomment-2110476094

PR #478, opened 2024-05-31
 bump k8s to 1.30

ericpromislow:
Related to issue https://github.com/rancher/rancher/issues/45339


alexandreLamarre (approved):

alexandreLamarre (approved):

alexandreLamarre (approved):

.github/renovate.json
```diff
@@ -12,7 +12,7 @@
     "github.com/rancher/hull",
     "github.com/rancher/lasso",
     "github.com/rancher/wrangler",
-    "github.com/rancher/wrangler/v2",
```
mallardduck, line 15: Unless I'm wrong - because Renovate config here applies to the whole repo we want to keep v2 listed for the Rancher 2.8 branch of BRO. Also this change would create 2 instances of V3 wrangler.

mallardduck (changes requested):
Just one small thing.

I also was unsure about the goreleaser changes too - guess we don't really need to publish windows or mac binaries so that's fine.

.github/renovate.json
ericpromislow: We probably want both `v2` and `v3` here because these list dependencies to ignore.

mallardduck (approved):
LGTM

joshmeranda (approved):
LGTM

PR #481, opened 2024-06-21
Add GHA comment commands for porting

mallardduck:
These GHA commands allow us to create comments like:

- `/backport release/4.0`
- `/forwardport release/5.0`
- `/backport v2.8.next1`

Where when it's on a PR you should provide a specific branch on this repo and it will create a new PR based on it targeting the new branch. And when it's an Issue it will create a new issue to track porting it to the tagged rancher release name.

### DIFF
--- a/.github/workflows/port-issue.yaml
+++ b/.github/workflows/port-issue.yaml
@@ -1,0 +1,103 @@
+# create a backport/forwardport of an issue when "/backport <milestone>" is commented
+name: Port issue
+run-name: "Port issue ${{ github.event.issue.number }}: ${{ github.event.issue.title }}"
+
+on:
+  issue_comment:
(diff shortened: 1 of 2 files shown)


mallardduck:
Would be neat to review and merge this one first - then test it on #480 since I want to see that backported to 2.8 too.

ericpromislow (approved):
Lots going on here, but these files are almost identical to the ones in `dashboard`, so approving

mallardduck:
Looks like I need to follow up here for 2 issues:

- Both workflows triggering resulting in 2 comments.
- Workflow cannot parse branches with `/` in the correctly;
  - errors about `release/v4.0` being parsed as `releasev4.0`

PR #488, opened 2024-07-17
goreleaser is failing with error 422 - this might fix it.

ericpromislow:


alexandreLamarre (approved):

mallardduck:
I don't think this will hurt, but it also seems like the fails are because it's trying to upload to the RC still. Since I'd you look at the RC release it has non-rc artifacts

mallardduck (changes requested):
I think instead of this we need to set `GORELEASER_CURRENT_TAG` to the tag we're building - that way when we have situations like right now where two tags point to the same SHA it won't get confused.

mallardduck:
Gonna close this for now since I think we solved it in #489

PR #490, opened 2024-07-19
[GHA] Reduce jobs on tag publish and fix tag var/context

mallardduck:
This PR does 2 important things:

1. Reduces the number of GHA jobs this repo spawns.
2. Fixes my last fix attempt properly.

Why those matter:

### Reducing GHA jobs
There are a lot of instances where varying rancher repos are failing due to odd rate-limit things and other quirks. Eventually these will be ironed out, but until then we can reduce the load we put on it by making this change. Since both the workflows rely on the same "pre-step" and call the CI workflow - that spawns 2 instances of CI with 3 jobs each, for 6 jobs total. And if one half of those fail/pass then the other should too (sans GHA worker issues).

### Fix to last attempt to multiple tags on identical commit
I misread the GHA docs and was looking at references for ENV based variables that are provided. However used an incorrect syntax and as a result the value was empty when reviewing the job. So that confirms my suspicion that the issue could have come back at next un-RC. By using the [`github` context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) though I can get the results I was going for with the first version of the fix.


alexandreLamarre (approved):

ericpromislow (approved):
Looks good, and stopped using `vars.GITHUB_REF_NAME`

PR #494, opened 2024-07-19
Adjust renovate groups & enable cron

mallardduck:
The current settings create really large PRs that mix all types of version bumps (major, minor, etc). And all the contexts (GH actions, docker files, go deps) into the same PR too.

With this change we will technically get more PRs but they will come thru grouped by what feel like safer groups. Since for example in this PR: https://github.com/rancher/backup-restore-operator/pull/493

I wouldn't feel as confident merging the `mergo` change from v0.3.16 -> v1.0.0. But do feel confident merging the GHA workflow changing from v5 -> v6. So this highlights where 2 major bumps are not equal (cause for concern).

This will allow us to more easily and simply just bump the GHA workflows as we need. Lettings us merge those CI bumps whenever we want. And then any actual bumps that directly go into the code we ship can be scrutinized without delaying those bumps.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,7 +1,7 @@
 {
   "extends": [
     "github>rancher/renovate-config#release",
-    "group:all"
+    "group:allNonMajor"
   ],
   "baseBranches": [
     "release/v3.0",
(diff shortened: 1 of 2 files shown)


mallardduck:
Here's an example of the PRs we'd expect to see when it runs after this config change:
Screenshot 2024-07-19 at 1 05 40 PM

ericpromislow (approved):
I don't follow everything that's going on here, but from the description it looks worth a try

PR #505, opened 2024-07-22
update codeowner to O&B instead of individual users

mallardduck:
This should adjust how the default reviewer works - so that it will assign it to our whole team instead of Eric everytime.


ericpromislow (approved):
LGTM, makes sense

Issue #483, opened 2024-07-03
[BUG][v2.9] Rancher Backups not persisting installed UI extensions


mallardduck:
**Rancher Server Setup**
- Rancher version: Any
- Kubernetes Version and Engine: N/A

**Describe the bug**
When using BRO to migrate Rancher to a new cluster the UI extensions do not persist.
This is because the CRD responsible for UI plugins are not being backed up.
Those are `uiplugins.catalog.cattle.io` and seem to be under `cattle-ui-plugin-system` namespace.

**To Reproduce**
1. Install a Rancher UI plugin,
2. Take a backup,
3. Migrate to a new cluster,
4. Observe the UI plugin is not active/installed.

**Expected behavior**
I believe we can expect these to be installed after a migration.
However we will want to work with UI team (anyone more expert on UI plugins than us) to understand this.
For instance, if UI plugins/extensions must be installed via clicking the UI buttons and applying CRDs won't work.

Note: For more context, we got word of the CRD needed via a [slack thread](https://suse.slack.com/archives/C02DG7NKU4R/p1720014888383319?thread_ts=1719954975.111299&cid=C02DG7NKU4R) (must be SUSE employee to see). Might be useful for other SME to reach out to.

jbiers:
Adding a UI extension to Rancher is done in two steps:

1. Fist the Admin has to enable extensions. This means installing the UI plugin operator chart, with all of its resources as listed [here](https://github.com/rancher/ui-plugin-operator/tree/main/charts/ui-plugin-operator/templates).
2. Then the desired extensions can be installed and will be persisted in the form of a `uiplugins.catalog.cattle.io` resource under the `cattle-ui-plugin-system` namespace.

Adding a new rule to the default ResourceSet to include `uiplugins.catalog.cattle.io` resources makes sure that these extensions will be included in a backup and can be restored into an empty cluster. However, this does not include the first step of **enabling extensions**. Once enabled manually, all of the installed extensions will be visible and ready to go.

It might also be possible to include backup rules for **all** of the resources created by the ui-plugin-operator chart, so the manual step of enabling extensions would be unnecessary. However, I'm not sure if this would be in the scope of this issue or even if this behavior should be expected, since enabling extensions is a very straightforward stateless 1-step task.

mallardduck:
This change can be found in:

- [v5.0.1-alpha.1](https://github.com/rancher/backup-restore-operator/releases/tag/v5.0.1-alpha.1)
[...]

PR #508, opened 2024-07-23
Add rule for UIPlugin resources in default ResourceSet

jbiers:
This PR seeks to resolve issue [#483](https://github.com/rancher/backup-restore-operator/issues/483)

## Test Steps
- In a local cluster with the Kamaji UI extension installed, manually edited the default ResourceSet to include the new rule for UIPlugins.
- Took a backup (B1)
- See if B1 correctly includes the Kamaji UIPlugin
- Create a new cluster and restore the backup as detailed [here](https://ranchermanager.docs.rancher.com/v2.6/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster)
- Once restored, enable extensions in the new cluster and reload the page. Check if Kamaji UI remains correctly installed.

### DIFF
--- a/charts/rancher-backup/files/default-resourceset-contents/rancher.yaml
+++ b/charts/rancher-backup/files/default-resourceset-contents/rancher.yaml
@@ -63,3 +63,7 @@
       - key: "resources.cattle.io/backup"
         operator: "In"
         values: ["true"]
+- apiVersion: catalog.cattle.io/v1
+  kindsRegexp: ^UIPlugin$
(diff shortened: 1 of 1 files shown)


mallardduck (approved):
LGTM; wonder if namespace part matters though.
Not sure we ever see any outside that namespace.
But maybe only gathering ones we know we use today is fine and if it changes later we add more. Rather than having to restrict things more later because we backup and restore too much?

PR #539, opened 2024-07-31
Revert "Update module dario.cat/mergo to v1 (release/v5.0)"

mallardduck:
Reverts rancher/backup-restore-operator#530

Per the note here: https://github.com/darccio/mergo?tab=readme-ov-file#100

> If the vanity URL is causing issues in your project due to a dependency pulling Mergo - it isn't a direct dependency in your project - it is recommended to use [replace](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive) to pin the version to the last one with the old import URL:
> 
> ```
> replace github.com/imdario/mergo => github.com/imdario/mergo v0.3.16
> ```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.22.0
 toolchain go1.22.5
 
 replace (
-	github.com/imdario/mergo => dario.cat/mergo v1.0.0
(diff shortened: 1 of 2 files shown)


ericpromislow (approved):
LGTM

PR #555, opened 2024-08-08
Add integration tests for arm64

dharmit:


mallardduck (changes requested):
I think once the OS fields have both values the action will work as expected. 👍

.github/workflows/ci.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
       matrix:
         os : [ 
           ubuntu-latest
-          # org-rancher-arm64-containers
+          # org-rancher--arm64-containers
```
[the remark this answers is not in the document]
dharmit: OK, I have made those changes. Let's see. 🤞🏽

.github/workflows/ci.yaml
```diff
@@ -60,9 +60,8 @@ jobs:
     runs-on : ${{ matrix.os }}
     strategy:
       matrix:
-        os : [ 
-          ubuntu-latest
-          # org-rancher-arm64-containers,
+        os : [
+          org--rancher--arm64-containers
```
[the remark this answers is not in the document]
dharmit: OK, I have made those changes. Let's see. 🤞🏽

.github/workflows/ci.yaml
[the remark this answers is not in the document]
dharmit: I made one mistake though. Forgot the extra `-` in the `build` step. Have fixed that now. Btw, how do you figure out that the queue for runners is long? I mean, is it visible on some status page or something like that?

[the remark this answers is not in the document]
mallardduck: I really wish there was a single page to check the queue - mostly I was going off intuition from seeing other `rancher/*` repos I'm working in having long queue times too. On each action that's queued it will tell you how long it's been waiting.

Sometimes if you're iterating rapidly and waiting on multiple layers of tests in the queue it's worth cancelling all but the newest test. That way you don't bother running CI for old commits if there is a newer one ahead of it that will be tested too. To do that normally I just navigate to the Action tab of the repo, then filter for "queued" status. And then carefully cancel only ones related to the PR you're working on. 😄

mallardduck:
Since i've just pushed our new `main` branch, I've repointed this PR to target `main`.
It should be a smooth transition since `main` is based on `release/v5.0`.

mallardduck:
Ok @dharmit - I put some more dots together here in this puzzle for us.
The labels we saw Alex working with weren't wrong, just 4 months old before our full adoption of GHA.
And things must have changed since then by a lot since our new docs look a bit different: https://github.com/rancherlabs/eio/wiki/GitHub-Actions-Runners

dharmit:
> And things must have changed since then by a lot since our new docs look a bit different: https://github.com/rancherlabs/eio/wiki/GitHub-Actions-Runners

They don't mention arm64 in containerized runners section. All occurrences of arm64 are under the "Non-Containerized Org-Level Runners (VMs and Bare Metal Instances)" section. 🤔

mallardduck:
I think we just need to try: https://github.com/rancherlabs/eio/wiki/GitHub-Actions-Runners#containerized-org-level-runners

And I suspect that they have arm ones ready to go?

.github/workflows/ci.yaml
```diff
@@ -29,8 +29,8 @@ jobs:
     strategy:
       matrix:
         os : [ 
-          ubuntu-latest
-          # org-rancher-arm64-containers
+          ubuntu-latest,
+          org--rancher--arm64-containers
```
mallardduck, line 33: So the matrix should become:
```
matrix:
  arch:
    - amd64
    - arm64
```

.github/workflows/ci.yaml
```diff
@@ -29,8 +29,8 @@ jobs:
...
           ]
     name : CI
     runs-on : ${{ matrix.os }}
```
mallardduck, line 36: This becomes:
```
org-${{ github.repository_owner_id }}-{{ arch }}-k8s
```

And then that should start matching the names of the runners we want to use:
`org-9343010-amd64-k8s`

mallardduck (changes requested):

.github/workflows/ci.yaml
```diff
@@ -26,14 +26,14 @@ on:
...
-        os : [ 
-          ubuntu-latest
-          # org-rancher-arm64-containers
-          ]
+        arch:
+          - amd64
+          - arm64
     name : CI
-    runs-on : ${{ matrix.os }}
+    runs-on : org-${{ github.repository_owner_id }}-{{ arch }}-k8s
```
mallardduck, line 36: `{{ arch }}` to `${{ matrix.arch }}` and I think we got it

mallardduck:
@dharmit - Disregard my suggestion to use a new CI file, I've worked with EIO to get this repo access to non-container runners. This provides us a much more direct drop-in replacement for the GHA runners. With those setup I was able to tweak this PR to use those runners for both arm and amd64 without major changes.

.github/workflows/ci.yaml
```diff
@@ -60,9 +59,9 @@ jobs:
     runs-on : ${{ matrix.os }}
```
mallardduck, line 59: Looks like we still need to update these other `matrix` references to use the new keys. So any `matrix.os` should end up being `matrix.arch`.

mallardduck (changes requested):
Still have some changes to be made before this works, I will leave these for you to resolve though now that we know the runners are sorted out and working. 😄

.github/workflows/ci.yaml
dharmit: I think we only need to change the name of the artifact to be unique for each arch. That's what I understood from https://github.com/actions/upload-artifact/issues/478 and have made a change accordingly. Let's see. 🤞🏽

.github/workflows/ci.yaml
```diff
@@ -60,9 +59,9 @@ jobs:
...
     strategy:
```
mallardduck, line 60: Will want to use the same matrix and runners as above down here too.

.github/workflows/ci.yaml
```diff
@@ -28,12 +28,11 @@ jobs:
...
-        os : [ 
-          ubuntu-latest
-          # org-rancher-arm64-containers
-          ]
+        arch:
+          - x64
+          - arm64
     name : CI
-    runs-on : ${{ matrix.os }}
+    runs-on : runs-on,image=ubuntu22-full-${{ matrix.arch }},runner=2cpu-linux-${{ matrix.arch }},run-id=${{ github.run_id }}
```
mallardduck, line 35: Also feel free to bump this to 4 CPU (should speed up build times)

.github/workflows/ci.yaml
```diff
@@ -49,20 +48,20 @@ jobs:
...
       build
     ]
-    runs-on : ${{ matrix.os }}
+    runs-on : runs-on,image=ubuntu22-full-${{ matrix.arch }},runner=4cpu-linux-${{ matrix.arch }},run-id=${{ github.run_id }}
     strategy:
       matrix:
-        os : [ 
-          ubuntu-latest
-          # org-rancher-arm64-containers,
+        os : [
```
mallardduck, line 62: replace OS with arch one:
```
        arch:
          - x64
          - arm64
```

mallardduck:
Might have to run workflows as debug to get additional logging. Looks like we're getting close but now seeing a new error related to no ARCH set and go binary missing.

dharmit:
> Looks like we're getting close but now seeing a new error related to no ARCH set and go binary missing.

I think the two are related because `ARCH` is set in `./scripts/version` as:
```sh
ARCH=$TARGET_ARCH
if [ -z "$ARCH" ]; then
    ARCH=$(go env GOHOSTARCH)
fi
```
Since go binary isn't available, both problems are popping up. Let me try to install go.

.github/workflows/scripts/install-go.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -e
+set -x
+
+GO_VERSION=1.22.0
+ARCH=`uname -i`
```
ericpromislow, line 7: Please use `uname -m` instead.  `-i` isn't supported on macOS and is documented as non-portable on ubuntu. As much as I can tell, they do the same thing.

.github/workflows/scripts/install-go.sh
```diff
@@ -0,0 +1,16 @@
...
+
+if [ $ARCH -eq "x86_64" ]; then
```
ericpromislow, line 9: It's clearer to set `PLATFORM=amd64|arm64` in a separate `case` block to avoid duplicating a long command.

But you should probably set this in `ci.yaml` and set the variables if they're unset so `.github/scripts/X` can still be run outside GHA, mostly for testing/dev purposes.

.github/workflows/scripts/install-go.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -e
```
ericpromislow, line 3: Instead do
```
set -eu
set -x
```

where the `set -x` could get dropped at a later date. And our scripts should always fail on unset vars

ericpromislow (changes requested):
I would drop the entire `install-go.sh` file and replace it by running `actions/setup-go` in `ci.yaml`

Also see how we deal with the two different ways of naming the amd architecture in `webhook/.../ci.yaml`.  And I would then try to pass a value for `ARCH` to `install-mc.sh` and for `TARGET_ARCH` for `./scripts/integration`

dharmit:
> I would drop the entire `install-go.sh` file and replace it by running `actions/setup-go` in `ci.yaml`

Thanks. I wasn't aware of it. 👍🏽 
 
> Also see how we deal with the two different ways of naming the amd architecture in `webhook/.../ci.yaml`. And I would then try to pass a value for `ARCH` to `install-mc.sh` and for `TARGET_ARCH` for `./scripts/integration`

I haven't made any changes that you suggested here. And, I think, it didn't cause any trouble because go installation happened before we run integration tests.

mallardduck (approved):

Issue #565, opened 2024-08-23
[CI] Release Branch Builds Specs


mallardduck:
## Possible branches:

- `main` - Our main dev branch, a default target for all changes; this represents the "next" version of BRO.
- `release/v[0-9]+.0` - Release specific branch correlated to a Rancher Minor version.
- `release/v[0-9]+.[0-9]+.[0-9]+` - A patch release specific branch; used rarely to create patch release of a specific minor.

> Note: the release branches above represent a format, not specific branches. However, only the listed branches, and one's matching the patterns, will be used to generate head builds.

### Release Tags

When a release is generated for each branch, the release will happen simultaneously on two tags. The first tag being a "static tag" matching the branch at that commit in time (i.e. `{branchName}-{commitSHA}`), the other a "rolling tag" matching the newest of that branch (i.e. `head`).

For the `main` branch we will generate a static tag of `{branchName}-{commitSHA}` and a rolling tag of `head`.   
On each release branch, we will generate a static tag like: `{branchName}-head-{commitSHA}` and a rolling tag of `{branchName}-head`.

Any branches outside of these will be ignored and not have automatic releases built.

### CI and Script changes

1. Review the existing: `scripts/version` - Can this be re-used/re-factored for our needs?
2. Add a new workflow just for `head-releases`; to not complicate the existing `publish.yaml` workflow.
3. Consider adding a new script to handle "Find previous release tag on current branch matching expected format"
4. Investigate best method to handle multiple tags being built (both goreleaser and container images)
5. Consider if we care about auto-changelogs on these head releases

mallardduck:
This is a tracking issue to follow issue we noticed here: https://github.com/rancher/backup-restore-operator/pull/562

This issue is follow up on this part:
> I will document for us a roadmap to get our CI fully embracing the new branching strategy. That way we will have proper head builds for release branches and main - built with both a static tag and a rolling `-head` tag.

mallardduck:
This will close out with the merge of these:

https://github.com/rancher/backup-restore-operator/pull/569
https://github.com/rancher/backup-restore-operator/pull/570
https://github.com/rancher/backup-restore-operator/pull/571

PR #566, opened 2024-08-29
[main] Fully Implement Head Tagging

mallardduck:
This PR adds scripts necessary for properly tagging the Image artifacts produced from "head builds". This way we can have an effective method to have a "rolling head tag" for each major branch without loosing access to past "head images".

This will close out https://github.com/rancher/backup-restore-operator/issues/565 with the only final goal being t he docker images as an artifact of head builds. As full releases with Charts and bins wouldn't really make sense since these are intended to improve Dev/QA workflows and not intended for other consumption.

---

For review of the action changes, please see: https://github.com/mallardduck/backup-restore-operator/actions/runs/10621704184

As that was a test run (of these same effective changes) before I prepared this branch for PR.

### DIFF
--- a/.github/workflows/head-releases.yaml
+++ b/.github/workflows/head-releases.yaml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - gha-testtesttest
       - release/v[0-9]+.0
 
 concurrency:
--- a/.github/workflows/head-releases.yaml
+++ b/.github/workflows/head-releases.yaml
@@ -1,0 +1,82 @@
+name : Branch Head Prerelease Images & artifacts (via goreleaser)
+
+on:
+  push:
+    branches:
+      - main
+      - gha-testtesttest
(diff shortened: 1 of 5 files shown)


mallardduck:
Once merged, this along with changes from the PR @dharmit made https://github.com/rancher/backup-restore-operator/pull/555 should be backported to release branches.

.github/workflows/head-releases.yaml
```diff
@@ -0,0 +1,82 @@
+name : Branch Head Prerelease Images & artifacts (via goreleaser)
+
+on:
+  push:
+    branches:
+      - main
+      - gha-testtesttest
```
jbiers, line 7: Do we want to keep the `gha-testtesttest` branch here?

mallardduck: No! Great catch - ty.

jbiers (approved):

Issue #482, opened 2024-06-26
Rancher restore fails with FleetWorkspaces


Daemonslayer2048:
**Rancher Server Setup**
- Rancher version: 2.8.1
- Installation option (Docker install/Helm Chart): Helm Install
- Kubernetes Version and Engine: v1.25.4+rke2r1

**Describe the bug**
When using FleetWorkspaces in Rancher this will create a new namespace for said workspace. Due to this if a user attempts to restore on a new cluster the restore process will fail as it will not create said namespace. 

**To Reproduce**
1. Create a fresh cluster 
2. Create FleetWorkspaces (See additional context below)
3. Install backup operator and take a backup
4. Delete cluster
5. Restore Rancher on totally new cluster
6. Observe restore failure

**Expected behavior**
I would expect one of two things to happen:   
1. Rancher restore should create the namespace as needed to allow the FleetWorkspace to be repopulated
2. Rancher restore will skip creating FleetWorkspaces as to not prevent the restore from completing 
Option one is preferable but two will at leas prevent end users from getting stuck. 

**Screenshots**
Not needed

**Additional context** 
Sample Fleet config
```yaml
---
apiVersion: management.cattle.io/v3
kind: FleetWorkspace
metadata:
  name: enterprise
 
---
apiVersion: management.cattle.io/v3
kind: FleetWorkspace
metadata:
  name: edge

---
apiVersion: fleet.cattle.io/v1alpha1
kind: ClusterGroup
metadata:
  name: infra
  namespace: enterprise
spec:
  selector:
    matchExpressions: []
    matchLabels:
      infra: "true"
---
apiVersion: fleet.cattle.io/v1alpha1
kind: ClusterGroup
metadata:
  name: apps
  namespace: edge
spec:
  selector:
    matchExpressions: []
    matchLabels:
      apps: "true"
```

ericpromislow:
Is this a dup of the other recent failure to save fleet secrets?

Daemonslayer2048:
Do you happen to know what issue number that would be?

ericpromislow:
https://github.com/rancher/rancher/issues/44033

dsmithbauer:
[...]

PR #576, opened 2024-09-10
Add FleetWorkspaces and their namespace to backup config

manno:
Refers to https://github.com/rancher/backup-restore-operator/issues/482

* add fleetworkspaces resource to backup config
* add any namespace created by that resource, identifiable by the "app.kubernetes.io/managed-by=rancher" label, too

The "fleet-local" and "fleet-default" workspaces are special, their namespaces don't have the label, but are backed up explicitly because their resource names start with "fleet-".

### DIFF
--- a/charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
+++ b/charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
@@ -1,6 +1,14 @@
+- apiVersion: "management.cattle.io/v3"
+  kindsRegexp: "^fleetworkspaces$"
 - apiVersion: "v1"
   kindsRegexp: "^namespaces$"
   resourceNameRegexp: "^fleet-"
+  # also match any namespaces created for workspaces
+  labelSelectors:
+    matchExpressions:
+      - key: "app.kubernetes.io/managed-by"
+        operator: "In"
+        values: ["rancher"]
 - apiVersion: "v1"
   kindsRegexp: "^secrets$"
   namespaceRegexp: "^cattle-fleet-|^fleet-"


charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
```diff
@@ -1,6 +1,14 @@
+- apiVersion: "management.cattle.io/v3"
+  kindsRegexp: "^fleetworkspaces$"
 - apiVersion: "v1"
   kindsRegexp: "^namespaces$"
   resourceNameRegexp: "^fleet-"
+  # also match any namespaces created for workspaces
+  labelSelectors:
```
mallardduck, line 7: This actually modifies the existing rule to only select `namespace` resources that start with `fleet-` prefix and (as of this change) only those that have `app.kubernetes.io/managed-by=rancher` labels.

I suspect based on the comment this should be a new rule like:
```yaml
- apiVersion: "v1"
  kindsRegexp: "^namespaces$"
  labelSelectors:
    matchExpressions:
      - key: "app.kubernetes.io/managed-by"
        operator: "In"
        values: ["rancher"]
```

This way the existing rule is unchanged and similar to the comment we will "also match for ns created for workspaces"

charts/rancher-backup/files/default-resourceset-contents/fleet.yaml
```diff
@@ -1,6 +1,14 @@
+- apiVersion: "management.cattle.io/v3"
```
mallardduck, line 1: This is already covered by:
https://github.com/rancher/backup-restore-operator/blob/f6209d69089f78de7653301ae3bb0b3fb54bd46d/charts/rancher-backup/files/default-resourceset-contents/fleet.yaml#L34

manno: Thanks, I didn't check for the group.

manno:
Closing this PR as discussed.

PR #606, opened 2024-10-30
Fix Elemental metadata backup

ldevulder:
Kind `metadata` is not saved, this PR should fix it. A backport to `release/v4.0` and `release/v5.0` is also needed.

### DIFF
--- a/charts/rancher-backup/files/default-resourceset-contents/elemental.yaml
+++ b/charts/rancher-backup/files/default-resourceset-contents/elemental.yaml
@@ -30,7 +30,7 @@
   resourceNameRegexp: "elemental.cattle.io$"
 - apiVersion: "elemental.cattle.io/v1beta1"
   kindsRegexp: "."
-  namespaceRegexp: "^cattle-fleet-|^fleet-"
+  namespaceRegexp: "^cattle-fleet-|^fleet-|^cattle-elemental-system$"
 - apiVersion: "rbac.authorization.k8s.io/v1"
   kindsRegexp: "^roles$|^rolebindings$"
   labelSelectors:


charts/rancher-backup/files/default-resourceset-contents/elemental.yaml
```diff
@@ -31,6 +31,10 @@
 - apiVersion: "elemental.cattle.io/v1beta1"
   kindsRegexp: "."
   namespaceRegexp: "^cattle-fleet-|^fleet-"
+- apiVersion: "elemental.cattle.io/v1beta1"
+  kindsRegexp: "^metadata$"
```
mallardduck, line 35: @ldevulder - Can these `metadata` resources exist in other namespaces?

I ask because the rule above this should capture it, if we add the relevant namespace (or prefix) to that one. So we could add `|^cattle-elemental-system$` - or some derivative that should match the resources to backup.

ldevulder: No it can only exist in `cattle-elemental-system`. But does this means that the `namespaces:` condition line added after `kindsRegexp` doesn't restrict the namespace?

mallardduck: Ok, cool. I was just curious as other Rancher sub-systems have required that type of prefix logic. If it's only ever a single specific namespace then my concern there is moot.

> does this means that the `namespaces:` condition line added after `kindsRegexp` doesn't restrict the namespace.

It does, but that's not what I was getting at - my point was that the rule directly above this is already intended to match for all of the resources matching CRD kinds for elemental. However because it only matches for specific namespace prefixes the `metadata` you're adding here is currently exclude.

In other words, because a rule already exists but isn't working as expected that rule is "bugged". So rather than adding a new rule, the first fix attempt would be to adjust the existing rule. Then only after dertemining if that's not possible we would add the new one.

---

All that said - the only reason to **not** edit the existing rule in favor of a new rule is if there are other Elemental resources within `cattle-elemental-system` other than `metadata` that we do not want to backup. If that's not the case then we're better off adding `|^cattle-elemental-system$` to line 33.

ldevulder: Ok,I understood. I will do a quick test locally and fix the regex if it's good.

mallardduck (approved):

Issue #612, opened 2024-10-31
[RFE] Add a cluster metadata mechanism for Backups


mallardduck:
# 🚀 Feature Request

This will (hopefully) further solve issues like this: https://github.com/rancher/rancher/issues/47581. 
It's not the first one like this we've gotten and even with this change may not be the last.  
However, until we try we can't say for sure that BRO itself can't make this more clear.  

## Background

When using BRO in the context of Rancher (i.e. Rancher Backups) the delineation between performing an "In-place Restore" and "Migration" can be very important. Currently the onus of understanding that is fully on the User interacting with the backups. This person needs to be well read on the Rancher Backups documentation, understand the state of the cluster when each backup point was created, and how that compares to the current cluster they're interacting with.

This is an easily overlooked aspect of how Rancher Backups is intended to operate. As issues around this come up we adjust documentation the best we can. Rather than continue to "patch via docs" we could implement a more foundational solution to serve as an indicator.

### Is your feature request related to a problem? Please describe.
Not exactly - more about clarifying a process that has some ambiguity.

### Describe the solution you'd like

- BRO team adds a mechanism for Backups to be correlated to the cluster they are created on.
- In turn, that is used to add a status field to the Backup indicating if it's valid for "in-place" restore.
- Finally, the Rancher UI (and CRD list) is updated to add a "In-Place Restore" column to display the status field in the Backups list.

Logically, the UI stuff needs to come last and can be considered a follow up effort. However we may even want/need to split the first two tasks into their own efforts/PRs.

We will need to explore the most effective way to get the results we're logically expecting. Since it's rather easy to dream up a metadata file we add to tars, but then that implies that to know if we can do an in-place we need to read the tar. Which won't scale well.

#### Technical Implementation

1. When backups are created add a new metadata field to contain cluster ID (of some sort),
2. Add a status field to the backup CRD based on if that metadata field matches the current cluster,
    - If the field is unset, status is unknown; or, false with a status condition might be better?
4. Add UI change to display that status field in table.

### Describe alternatives you've considered

While I have considered this idea in the past (a few times over the last few months), none of those ideas were viable as functional options nor worth documenting. At the current time, the only other alternative for solving this issue is adding more docs.

### Additional context

N/A

mallardduck:
[...]

PR #613, opened 2024-10-31
[main] Add new method to track originating cluster of Backup

mallardduck:
This partial solves for https://github.com/rancher/backup-restore-operator/issues/612

However it's important to note that while I wrote an RFE for this too - the root here is seeking to fix a bug. The bug is that "BRO itself doesn't do anything to give indications when In-Place Restore is acceptable to use".

The short version of the fix, is we add a few new status details and expose them via `kubectl`. And later make a change to `rancher/dashboard` to ensure UI can also benefit from this new context by adding a column.

---

To be discussed:

- This change is very Backwards Compatible, can we ship a CRD change for the new status field in a `2.10.x` release?
  - Feels like it's a grey area - but, probably OK.
  - `k8s` seems to be OK with full on schema additions (of `spec` too) as long as they're optional/nullalble. This is hopefully less impactful since users cannot control `status`.
  - The current result on a cluster w/o CRD change is:
    -  is a new warning about `originCluster` not existing on status,
    - Status conditions are correctly set to represent unknown in-place viability.
- Do we want to explore supporting an even more Backwards friendly method of detecting origin ID's with Filename?
  - For the best UX, we will account for existing backups with UID in filename. In other words, once the CRD is updated with new status field we'll add the UID we find from the filename (or do nothing if not matching expected filename format).

### DIFF
--- a/pkg/controllers/backup/cluster_origin.go
+++ b/pkg/controllers/backup/cluster_origin.go
@@ -16,7 +16,7 @@ type backupClusterOriginConditionMeta struct {
 	currentInPlaceRestoreCondition bool
 }
 
-func newBackupClusterOriginConditionMeta(controllerClusterId string, backup *v1.Backup) backupClusterOriginConditionMeta {
+func newBackupClusterOriginConditionMeta(controllerClusterID string, backup *v1.Backup) backupClusterOriginConditionMeta {
 	conditionMeta := backupClusterOriginConditionMeta{
 		backupName:                     backup.Name,
 		hasClusterOriginID:             false,
@@ -40,7 +40,7 @@ func newBackupClusterOriginConditionMeta(controllerClusterId string, backup *v1.
 	}
 
 	if conditionMeta.hasClusterOriginID {
-		conditionMeta.canInPlaceRestore = conditionMeta.clusterOriginID == controllerClusterId
+		conditionMeta.canInPlaceRestore = conditionMeta.clusterOriginID == controllerClusterID
 	}
 
 	currentInPlaceRestoreString := condition.Cond(v1.BackupConditionInPlaceRestore).GetStatus(backup)
--- a/pkg/controllers/backup/cluster_origin.go
+++ b/pkg/controllers/backup/cluster_origin.go
@@ -27,10 +27,10 @@ func newBackupClusterOriginConditionMeta(controllerClusterID string, backup *v1.
 		currentInPlaceRestoreCondition: false,
 	}
 
-	originAnnotationValue, ok := backup.GetAnnotations()[v1.BackupClusterOriginIndex]
-	conditionMeta.hasClusterOriginID = ok && originAnnotationValue != ""
+	originalValue := backup.Status.OriginCluster
+	conditionMeta.hasClusterOriginID = originalValue != ""
 	if conditionMeta.hasClusterOriginID {
-		conditionMeta.clusterOriginID = originAnnotationValue
+		conditionMeta.clusterOriginID = originalValue
 	}
 
 	currentOriginConditionString := condition.Cond(v1.BackupConditionClusterOrigin).GetStatus(backup)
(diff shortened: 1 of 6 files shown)


mallardduck:
@jbiers - So i'm not moving it to in review yet, since this was a very rough first pass try. However if you can take some time to get familiar with the changes and review the list of concerns I thought of (as well as consider/add your own) - that'd be great help to keep the ball rolling on this. In other words, while I could work on this more I'd love your feedback first before I spend more time on it.

pkg/util/util.go
```diff
@@ -74,3 +78,47 @@ func ErrList(e []error) error {
...
+		logrus.Infof("Error fetching CRD: %v", err)
+		return false
+	}
+
+	// Inspect the status schema, for example
+	_, found := crd.Spec.Versions[0].Schema.OpenAPIV3Schema.Properties["status"].Properties["originCluster"]
+	if found {
+		logrus.Debugf("Status schema contains `originCluster` on CRD `%s`.\n", crdName)
+		return true
+	}
```
alexandreLamarre, line 120: Going back to the above discussion, storing the cluster metadata in the backup CRD spec as a nullable obejct shouldn't be a breaking change, and we can instead check if it is nil or not, rather than doing dynamic CRD reflection

IMO it is easier to make it testable if we avoid checking crd definitions and focus on the presence/absence of a new field to determine that.

pkg/controllers/backup/cluster_origin.go
```diff
@@ -0,0 +1,109 @@
...
+	conditionMeta.hasInPlaceRestoreCondition = currentInPlaceRestoreString != ""
+	if !conditionMeta.hasInPlaceRestoreCondition {
+		conditionMeta.currentInPlaceRestoreCondition = currentInPlaceRestoreString == "True"
+	}
+
+	return conditionMeta
+}
+
+// prepareClusterOriginConditions helps set the cluster origin conditions and reports if anything changed in this part of status.
+func (h *handler) prepareClusterOriginConditions(backup *v1.Backup) bool {
```
alexandreLamarre, line 56: I think a lot of this logic in this function can also be simplified/more readable if we encapsulate most of it in struct methods

pkg/controllers/backup/controller.go
```diff
@@ -39,7 +39,9 @@ type handler struct {
 	dynamicClient           dynamic.Interface
 	defaultBackupMountPath  string
 	defaultS3BackupLocation *v1.S3ObjectStore
-	kubeSystemNS            string
+	// TODO: rename to kubeSystemNamespaceUID; nit to improve clarity, it's not the string representation nor the NS resource
+	kubeSystemNS              string
+	canUseClusterOriginStatus bool
```
alexandreLamarre, line 44: stylistically, should avoid storing additional boolean fields, in favor of a method. 

I think this field might be useless?

If we opt into using the backup CRD spec for storing cluster metadata, and older backup-restore versions don't set this new null-able field, then we cannot determine where it is from.

pkg/controllers/backup/cluster_origin.go
```diff
@@ -0,0 +1,109 @@
...
+	conditionChanged := false
+	if !h.canUseClusterOriginStatus {
+		currentOriginConditionString := condition.Cond(v1.BackupConditionClusterOrigin).GetStatus(backup)
+		if currentOriginConditionString != "False" {
+			condition.Cond(v1.BackupConditionClusterOrigin).SetStatusBool(backup, false)
+			condition.Cond(v1.BackupConditionClusterOrigin).Message(backup, "CRD not updated to include cluster UID yet.")
```
alexandreLamarre, line 62: When would this realistically happen?

alexandreLamarre (commented):
Mostly some small comments about making the code more simple,readable & testable.

Since this PR could be a stepping stone to fully allowing backups of backup CRDs themselves, storing the cluster origin metadata in the Status feels wrong.

But if we do intend to fully support taking backups of backup CRDs, then even storing the cluster origin metadata in the Spec section of the CRD feels wrong.

What happens when we have a list of remote backups say `A`, `B`, `C`, all taken from different origins?


If the current cluster matches origin `C` then our status will indicate the restore is viable even if they specify an older backup `A` or `B`. I know this hypothetical would be rare but I think it points to the fact that maybe we should think about this differently.

The current design labels the recurring backups as coming from this cluster only, whereas specific backups are specific to specific clusters

What if we embedded cluster origin metadata not in the CRDs, but the backup data itself? 

This way the flow looks like 
- Each remote backup contains a cluster origin
- Each restore pointing to a specific backup can determine before actually restoring objects if it is viable
-  We can add a flag to the restore spec to allow `unsafe` backups, ones where the cluster origin doesn't match
- By default, we disallow cases where we could have unsafe restores
- If the restore is unsafe, the status is surfaced through the restore CR
- Users can then modify the restore CR to restore the backup even if it is unsafe after the fact, or opt in before creating the restore CR


From the end user's perspective it is somewhat clunkier to take a restore, then be informed that it is potentially unsafe, but as an end user I value being shown correct information more


Additionally storing, the cluster origin metadata in the backup data, allows us to add tests for each corner case in the e2e tests when taking a restore, producing a potentially less volatile new feature

pkg/controllers/backup/cluster_origin.go
mallardduck: Yeah, at least I think so. Granted it was during devworkflows, but I did run into it myself.

Maybe not as big a concern in prod, but I think could still happen - in a chart like BRO if they don't also update the CRDs. I.e. Installed via helm CLI directly.
Or even if (an edge case) where a user manually targets a newer image than the chart. However if we opt to not change the CRD to add it to status this wouldn't be needed anyway.

jbiers:
While testing some stuff with our backups I found something we may not have taken into account: if you apply a One-Time backup in a different cluster, the operator will actually pick it up and execute it again. For that reason, it in-place backups will always be possible

alexandreLamarre:
:/

jbiers:
:O

PR #625, opened 2024-12-05
omit DWARF debug info when building BRO binaries

alexandreLamarre:


mallardduck:
Will want to have this target `main` and then you can use `/backport {branch}` to have GHA create backport PRs. Or you can manually create backport PRs too obviously.

mallardduck:
Hmm. Odd - looks like the backport action failed because it doesn't identify you as a member of rancher org. 🤔 Seems odd, maybe an Okta profile issue? AFAIK if our SUSE okta has our GitHub username in the profile it will auto add/invite us to Rancher org.

Also BTW @jbiers and I opted to start using `6.x` for the branch name since it's more accurate than using zero.

/backport release/v6.x

mallardduck (approved):

mallardduck:
@alexandreLamarre - i still might need to fix the comment part since that didn't work here for some reason. (tho it did work in my testing before merging the PR to fix it.) however the backport command should work as intended now. 

At least barring [permission issues](https://github.com/rancher/backup-restore-operator/actions/runs/12281322417?pr=625+%28comment%29):

> alexandreLamarre is not a member of rancher

As mentioned you may just need to adjust Okta settings to ensure your user shows up as in the Rancher org - if you are in the org, I think your membership needs to be public for it work. So that could be a factor too if you're in the org currently. If possible maybe we can create a workaround for that to allow that membership to stay private and still be identifiable.

Issue #561, opened 2024-08-13
[CI] Build/Test Improvement suggestions from EIO


mallardduck:
- [x] - It looks like you could probably do a go mod download prior to tests/builds, if you wanted to cache that.
- [ ] - 76sec to build the binary, might be throwing a few more CPU cores at it to see how much it cuts down?
- [x] - Not performance related, but you may want to consider adding -s to your GOLDFLAGS (which doesn’t break stack traces, unlike post-processing it with strip can do), and possible -w (unless the app uses or exposes an interface for pprof profiling info). Both of those can cut down your binary size, and smaller containers are better for everyone.

NOTE: just for long term tracking and memory here. Not a call to action.

alexandreLamarre:
https://github.com/rancher/backup-restore-operator/blob/b72bef8c95a62f4a49f6e7015a2cd71bf40b39d4/package/Dockerfile#L9

`go mod download` was one of the first optimizations made for building images when we switched to github actions, unless im missing something here.

The `-w` flag looks mistakenly omitted again during the GHA migration. But the -s flag has been present since then. 

We can look into adding more cpu cores to speed up the builds, but the culprit here behind slow build times is reliance on dapper for CI. This is a bigger tech debt item to address, but we can certainly add it to this issue.

alexandreLamarre:
Adding -w flag : https://github.com/rancher/backup-restore-operator/pull/625

mallardduck:
Know this was closed as stale, but @alexandreLamarre essentially did the work to solve this one.

PR #626, opened 2024-12-11
Cleanup dapper and code

alexandreLamarre:
Part of : https://github.com/rancher/backup-restore-operator/issues/561


pkg/objectstore/s3minio.go
```diff
@@ -135,7 +146,7 @@ func GetS3Client(ctx context.Context, objectStore *v1.S3ObjectStore, dynamicClie
 		log.Tracef("Found secretKey [%s] in secret [%s] in namespace [%s]", secretKey, secretName, secretNs)
 	}
 	// if no s3 credentials are provided, use IAM profile, this means passing empty access and secret keys to the SetS3Service call
-	s3Client, err := SetS3Service(objectStore, accessKey, secretKey, true)
+	s3Client, err := SetS3Service(objectStore, accessKey, secretKey, false)
```
alexandreLamarre, line 149: Need to inject an additional option when storing secrets to set SSL to false, but always safeguard to defaulting to true unless explicitly opted out of

.github/workflows/ci.yaml
alexandreLamarre, line 1: Need to update goreleaser and release CI as well

scripts/package
mallardduck, line 1: Seems there's an existing bug with this RE image tags.
The issue comes from the `script/version` file though.

The tag appended to the image will have a suffix set based on the CPU arch. So even when you opt to use buildx - which currently produces only amd64 but could really produce a multiarch - will have the CPU suffix of the native arch, not the target arch.

I think a simple fix is just to fully remove that aspect of the scripts. Since it's not needed for real image releases (as those directly build the `Dockerfile`), nor is it really helpful for dev usage IMO.

alexandreLamarre: yeah good catch hadn't paid much attention to that part

pkg/resourcesets/collector.go
```diff
@@ -110,13 +110,13 @@ func (h *ResourceHandler) gatherResourcesForGroupVersion(filter v1.ResourceSelec
...
 	groupVersion := filter.APIVersion
-	logrus.Infof("Gathering resources for groupVersion: %v", groupVersion)
+	logrus.Debugf("Gathering resources for groupVersion: %v", groupVersion)
 
 	// first list all resources for given groupversion using discovery API
 	resources, err := h.DiscoveryClient.ServerResourcesForGroupVersion(groupVersion)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			logrus.Infof("No resources found for groupVersion %v, skipping it", groupVersion)
+			logrus.Debugf("No resources found for groupVersion %v, skipping it", groupVersion)
```
alexandreLamarre, line 119: This seemed very verbose for Info logging, so I set them to Debug. WDYT?

pkg/util/initializer.go
```diff
@@ -0,0 +1,60 @@
+package util
+
+import (
+	"context"
+	"sync"
+)
+
+type Initializer struct {
```
mallardduck, line 8: Not sure I'm following everything going on here - but I think I get the gist.
A new mechanism to control global state related to controller config/settings.

.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
```
mallardduck, line 1: Maybe want an exclude for this? I use goland and exclude my `.idea` folder.

mallardduck (approved):

mallardduck (changes requested):
err actually sed still an issue for the helm command:
```
+ sed -i -e 's/^version:.*/version: 26859b8/' -e 's/^appVersion:.*/appVersion: 26859b8/' -e 's/catalog.cattle.io\/upstream-version:.*/catalog.cattle.io\/upstream-version: 26859b8/' build/charts/rancher-backup/Chart.yaml
sed: -e: No such file or directory
make: *** [package-helm] Error 1
```

Current PR change set fixes this, GG!

.vscode/launch.json
alexandreLamarre: Left it in so others can use debug configurations out of the box, but I don't have a strong opinion either way

pkg/util/initializer.go
alexandreLamarre: TLDR; its a synchronization primitive wrapper that ensures its closure is only set once, and anyone that shows interest in what it encapsulates must wait for it to be set. It was used a lot in opni

mallardduck (approved):
Seems the macos `sed` issue is resolve now.
Additionally I verified that `make` (default) will work for me too.

It did have some initial issues, but those were all local caused. Cleaning up the repo to remove old files from `dist` and `build` seemed to fix that.

jbiers (commented):
LGTM

jbiers (dismissed):

mallardduck (approved):

Issue #607, opened 2024-10-30
[RFE] Provide 2 default ResourceSet configs out of the box based on use-case


mallardduck:
# 🚀 Feature Request

## Background

Historically Rancher Backups (BRO) has opted to exclude sensitive data from backups by default as it's seen as a security risk. This in turn can lead to some less than stellar experience for customers using BRO from Backups or Migrations. Of particular concern are areas related to Fleet, Provisioning, and cluster config related credentials.

In the more recent era of BRO we've faced far too many support issues to continue with the philosophy of excluding sensitive data from Backups. As that ultimately leads to support cases during some extremely tense situations. To balance these concerns we've gradually shifted focus to push users to utilize the encryption features.

With [the recent change](https://github.com/rancher/charts/pull/4586) on this feature we fully support the same Encryption options k8s does. This opens the door for us to adjust how we handle Secrets in our default ResourceSet and to take our efforts to suggest encrypting backups even further.

### Is your feature request related to a problem? Please describe.
Yes, there are a number of open issues this is related to.

Put simply, this RFE will help address many issues that are related to BRO capturing or skipping various sensitive data (Secrets, etc). In the past BRO opted to intentionally not capture sensitive data to avoid potential risk of those secrets being accessible via the Backup file. Recent changes to BRO allow more complex encryption configs to be used which provides the opportunity for us to include and encrypt sensitive data in the backups.

### Describe the solution you'd like
There are two overlapping concepts that this issue relates to within BRO, that is:

- Ability to Encrypt BRO backups at rest, and
- the rules in the default `ResourceSet` that BRO provides.

Using the functionality provided by the new Encryption features we can ship 2 default `ResourceSet` versions; one matching our existing one, and one that's meant to capture all the necessary secrets for Rancher to work. In our documentation and in our examples, we will need to compare and contrast the pros/cons of each; these mainly boil down to:

- On the "Basic" `ResourceSet`, the backup files will not contain any sensitive or secrets data and not need encryption to be suggested/required to use. In this use-case the secrets necessary for the cluster which cannot/will-not be recreated should be backed up outside of BRO, or otherwise configured to deploy via CI/CD.
- On the "Full" `ResourceSet`, the backup files will include any necessary secrets to fully restore Rancher to a working state. Any examples and documentation for this mode must also show usage of an EncryptionConfiguation.
[...]

PR #634, opened 2024-12-30
Create resourceset alternative

jbiers:
# Issue: https://github.com/rancher/backup-restore-operator/issues/607

This PR implements two new ResourceSets for backing up Rancher. These are *basic* and *full*. To prevent this from being sudden a breaking change, the old *rancher-resource-set* will be kept for now but be handled as deprecated. A follow-up issue will be opened to remove it completely in the future, possibly in 2.12 release.


examples/encryption-provider-config.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apiserver.config.k8s.io/v1
 kind: EncryptionConfiguration
 resources:
   - resources:
-      - secrets
+      - "*.*"
```
alexandreLamarre, line 5: Is there a reason our example encryption includes all resources now? There's usually a performance overhead to encrypting all resources

mallardduck: Agreed here - I think we can leave the example to only encrypt the sensitive data.

However we could include a note, in the docs update half of this effort, that communicates our users COULD opt to encrypt everything if they want to incur the extra cost (time/disk space) to encrypt all the manifests.

mallardduck: If we want this "encrypt all things" example for testing needs, that's fine. But lets make a new manifest for that example.

jbiers: Agree to have this as a documentation note. Reverted back to only encrypting secrets.

charts/rancher-backup/templates/rancher-resourceset-full.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: ResourceSet
+metadata:
+  name: rancher-resource-set-full
+controllerReferences:
+  - apiVersion: "apps/v1"
+    resource: "deployments"
+    name: "rancher"
+    namespace: "cattle-system"
+resourceSelectors:
+{{- range $path, $_ := .Files.Glob "files/basic-resourceset-contents/*.yaml" -}}
```
mallardduck, line 11: NGL, I didn't think this far ahead as far as "how we maintain two rule sets w/o redundant rules". I really like this as the solution overall as it fits our needs perfectly right now.

---
Down the road we may want to consider something different if it gets complex feeling to maintain it this way.

But IMHO that should be an effort of its own outside of this. As we would want to evaluate if we can create a sort of "Intermediate" structure to represent these rules. It would need to be something that allows tracking more additional structured data about the rules and then processes them into the CRD supported rule syntax. Certainly an existing area to imagine improving later but not critical in anyway today.

jbiers: Yep, agree there's room for improvement here in the future. Maybe we can open a separate issue for tracking this and dumping our ideas until we have something concrete to be implemented. WDYT?

mallardduck: Yeah, let's just open a tech-debt one but not make it a priority (so pre-set it as in backlog). Something sorta generic so we can collect ideas there, like "New build system ides for ResourceSet rules". Then as we work on other BRO issues that are already a priority, if we run into ideas on how to improve that area we can just drop in notes and work up a solution when we feel ready for that.

mallardduck (approved):
LGTM.

PR #644, opened 2025-01-14
Updates & Refactors for k8s 1.32

mallardduck:


pkg/resourcesets/collector.go
```diff
@@ -436,17 +436,21 @@ func writeToBackup(ctx context.Context, resource map[string]interface{}, backupP
...
 
-		resourceBytes, err = json.Marshal(encrypted)
+	// Since k8s 1.32 we cannot verify a transformer must be run, so it's always run now.
+	maybeEncrypted, err := transformer.TransformToStorage(ctx, resourceBytes, value.DefaultContext(additionalAuthenticatedData))
+	if err != nil {
+		return fmt.Errorf("error converting resource to JSON: %v", err)
+	}
+
+	// Need to verify encrypted is actually different...
+	if !bytes.Equal(resourceBytes, maybeEncrypted) {
```
mallardduck, line 447: Because we can do this check here...

pkg/resourcesets/collector.go
```diff
@@ -436,17 +436,21 @@ func writeToBackup(ctx context.Context, resource map[string]interface{}, backupP
 	if err != nil {
 		return fmt.Errorf("error converting resource to JSON: %v", err)
 	}
-	if transformer != nil && !util.IsDefaultEncryptionTransformer(transformer) {
```
mallardduck, line 439: ...it technically does similar to what this intended to do, just after `transformer.TransformToStorage` is called instead of before. Obviously before was better, but no longer an option (for) now.

pkg/resourcesets/collector.go
```diff
@@ -436,17 +436,21 @@ func writeToBackup(ctx context.Context, resource map[string]interface{}, backupP
...
-		encrypted, err := transformer.TransformToStorage(ctx, resourceBytes, value.DefaultContext(additionalAuthenticatedData))
-		if err != nil {
-			return fmt.Errorf("error converting resource to JSON: %v", err)
-		}
 
-		resourceBytes, err = json.Marshal(encrypted)
+	// Since k8s 1.32 we cannot verify a transformer must be run, so it's always run now.
+	maybeEncrypted, err := transformer.TransformToStorage(ctx, resourceBytes, value.DefaultContext(additionalAuthenticatedData))
```
mallardduck, line 441: I think that because of those factors, this is a new change in dataflow/processing that non-encrypted restores used to skip, but will now run all the time. So ultimately this likely doesn't change too much at the end of the day, but is technically a new area that an error could come from.

alexandreLamarre (commented):
We should bump
https://github.com/rancher/backup-restore-operator/blob/79a0a70cb908e645d34d3410b54e92dfd2fb0039/.github/workflows/ci.yaml#L72

The matrix k3s versions here to 1.30.X & 1.32.X

alexandreLamarre (approved):

alexandreLamarre:
Part of https://github.com/rancher/rancher/issues/48832

PR #645, opened 2025-01-14
Refactor Encryption logic for improved testing

mallardduck:
Before the change there are 2  files in `util`; 0% and 22% respectively.
Before this change, the `util` package is currently at 50% coverage.

After the change there are 3 files (counting new sub package); 0%, 13% and 70.4% respectively.
After this change, the `util` package will be at 66.7% coverage.


pkg/util/encryptionconfig/config_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+var deploymentGVR = schema.GroupVersionResource{
+	Group:    "apps",
+	Version:  "v1",
+	Resource: "deployments",
+}
+
+func TestGetEncryptionTransformersFromSecret_Basic(t *testing.T) {
+	encryptionConfigFilepath := filepath.Join("testdata", "encryption-provider-config-wildcard.yaml")
+	configBytes, _ := os.ReadFile(encryptionConfigFilepath)
```
alexandreLamarre, line 30: Instead of the `filepath` + `os.Read` can use:
 https://github.com/rancher/backup-restore-operator/blob/155b1a97b134e43fdb88130c898f0d72690365a1/e2e/test/test.go#L14
 
 this way the tests aren't dependent on specific paths
 
 can also probably put move that into a `pkg/test` instead of `e2e/test` if you want

alexandreLamarre (approved):

pkg/util/encryptionconfig/config_test.go
mallardduck: Maybe as a follow up PR we move that to test, and consider making it more generic? To use:
```
func Data(testFs embed.FS, filename string) []byte {
```

So then we can still keep organizing each pkgs test stubs in place?

mallardduck:
Manual backporting it is...

pkg/util/encryptionconfig/config.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Open the file for reading (or other operations) and return the handle
+	file, err := os.Open(EncryptionProviderConfigKey)
+	if err != nil {
+		return nil, err
+	}
+
```
alexandreLamarre, line 63: I didn't notice this before, but you we should be careful about opening files inside a function and expecting others to close said file. 

Can lead to concurrent closes or file descriptor leaks. 
We should look into alternative return types instead of `os.File`. I also don't see this being used by anything other than the tests at the moment. 

Will we need this inside the controller logic?

alexandreLamarre:
Part of https://github.com/rancher/rancher/issues/48832

Issue #352, opened 2023-05-08
Add prometheus metrics to backups


erdeerdeerde:
Hello, I would like to monitor the backups with prometheus. 
As far as i could see there are no metrics available.

Specifically i would be interested in:
- When was the last backup executed.
- Was the last backup successful?
- What is the size of the last backup.

Would it be possible to add those metrics? 

Thanks for your great work!

gcezaralmeida:
Hi. Any news about this feature?

David-VTUK:
@gcezaralmeida - I've added Backup/Restore metrics to https://github.com/David-VTUK/prometheus-rancher-exporter if you're willing to try it and provide feedback that would be great!

Example dashboard:

![image](https://github.com/rancher/backup-restore-operator/assets/5892615/e98e51bb-4e80-4171-bd99-5035e73cb040)

(I'm a Suse employee but work on this as a side project in an unofficial capacity.)

mallardduck:
IMO we could reasonably aim for 2.10.1 or 2.10.2 on this one.

hansbogert:
There should be a way to alert if a backup fails. Seems logical to include this right?

jbiers:
## Notes for QA testing

To properly test this, first have rancher-monitoring installed.

This issue includes these fields in the rancher-backup chart:
```monitoring:
  metrics:
    enabled: false

    serviceMonitor:
      enabled: false
  
      additionalLabels: {}
      metricRelabelings: []
      relabelings: []
```

Set both the metrics and the serviceMonitor to `enabled: true` and you should see a serviceMonitor and a regular service created in the release namespace (default being cattle-resources-system). Check that both exist. Also the service should have these annotations:

```
    prometheus.io/path: /metrics
    prometheus.io/port: /8080
    prometheus.io/scrape: "true"
```

Once those are deployed, Prometheus should automatically pick up rancher-backup as a target. This can take a couple minutes but.

Once rancher-backup has added as a Prometheus target, you can query the metrics from Prometheus. The metrics are defined [here](https://github.com/rancher/backup-restore-operator/blob/main/pkg/monitoring/metrics.go#L21), and you should be able to query them all.

Next, you can create a regular Backup and see if the metrics reflect it correctly:
- _rancher_backup_ should have the correct details for this backup
- _rancher_backup_count_ should return one
- _rancher_backups_attempted_ should return one
- _rancher_backups_failed_ should return zero

[...]

PR #648, opened 2025-01-17
Add prometheus metrics

jbiers:
## Issue: 
Solves [#353](https://github.com/rancher/backup-restore-operator/issues/352). Also relates to SURE-8367.


charts/rancher-backup/values.yaml
```diff
@@ -38,7 +38,7 @@ persistence:
   size: 2Gi
 
 # Add log level flags to backup-restore
-debug: false
+debug: true
```
alexandreLamarre, line 41: should remember to reset this to false when the PR is ready

pkg/controllers/backup/controller.go
```diff
@@ -85,12 +89,24 @@ func Register(
...
+
+	if h.metricsServerEnabled {
+		backupList, err := h.backups.List(k8sv1.ListOptions{})
+		if err != nil {
+			logrus.Error("Error getting Backup CR list. Failed to update metrics.")
+			return nil, err
+		}
+
+		monitoring.UpdateBackupMetrics(backupList.Items)
+	}
```
alexandreLamarre, line 102: This should run as a background task, instead of when we enqueue a backup resource

pkg/monitoring/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+		prometheus.GaugeOpts{
+			Name: "bro_backup",
+			Help: "Details on a specific Rancher Backup CR",
+			// add labels showing encryption type and current status
+		}, []string{"name", "resourceSetName", "retentionCount", "backupType", "filename", "storageLocation", "nextSnapshot", "lastSnapshot"})
+
+	backupLastProcessed = promauto.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "bro_backup_last_processed",
+			Help: "Unix time of when the last Backup was processed (in seconds)",
```
alexandreLamarre, line 31: Metric names should encode the units per naming conventions https://prometheus.io/docs/practices/naming/

pkg/monitoring/metrics.go
alexandreLamarre, line 1: I think useful metrics to have are:
- bro_backups_attempted, counter (labels : name)
- bro_backups_failed, counter (labels : name)
- backup_duration_ms (histogram) (labels: name)

pkg/monitoring/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+		})
+
+	backupFailed = promauto.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "bro_backup_failed",
+			Help: "Indicates whether the last Backup to be processed failed or not",
+		})
```
alexandreLamarre, line 38: This is a good idea, but in practice counters with the total number of failures are more useful, not just the latest process

pkg/controllers/restore/controller.go
```diff
@@ -126,6 +130,17 @@ func Register(
...
 func (h *handler) OnRestoreChange(_ string, restore *v1.Restore) (*v1.Restore, error) {
+
+	if h.metricsServerEnabled {
+		restoreList, err := h.restores.List(k8sv1.ListOptions{})
+		if err != nil {
+			logrus.Error("Error getting Restore CR list. Failed to update metrics.")
+			return nil, err
+		}
+
+		monitoring.UpdateRestoreMetrics(restoreList)
```
alexandreLamarre, line 141: Similarly, should be run in a background loop, and suggest using backoffs for apierrors

charts/rancher-backup/templates/service-monitor.yaml
alexandreLamarre, line 1: If I recall, one of the original asks was to add alerting rules along with the metrics as well, but we can always follow up in a later PR with that

jbiers: It wasn't in the original request, but adding alerting rules is something I absolutely want to to. The plan is a later PR :D

pkg/monitoring/metrics.go
jbiers: This was a request from the user who created the original issue. Not sure if it's still relevant though. A metric keeping track of the total number of failures is interesting regardless of that.

pkg/monitoring/metrics.go
jbiers: You mean these as the only metrics to be added to the backups? Or as extras to the existing ones? I ask this question since the user on the original issue had some other specific requests which would be useful to him and, I suppose, could be of value to other users as well.

jbiers: Also, how would *bro_backups_attempted* be different from the existing *bro_backup_count*? Or is this more about naming?

jbiers: As this was a very crude first pass on implementing metrics for BRO, suggestions on potentially useful metrics could be useful. Pinging @mallardduck here for his opinion as well since it's mostly us three working on the project.

alexandreLamarre: bro_backup_attempted would be a monotonic increasing counter, which would increment each time a backup task is run. For example in the case of recurring backups, you would have many attempted backups and many failed backups

pkg/monitoring/metrics.go
alexandreLamarre: Metrics typically aggregate over statuses or conditions, so it seems like an anti-pattern to have a metadata metric keeping track of the last backup that has failed which is stateful (and could flip between 2 different recurring backups which makes it less insightful), rather than a record of all failed backups that can be queried, aggregated and analyzed

alexandreLamarre: The ask of the last backup failed is implicit in keeping track of all failures, and that's how metric exporters work

alexandreLamarre:
Also @jbiers need to update the rancher/hull tests for the new chart flag

jbiers:
Next steps:
- Refine and collect Restore metrics.
- Create useful prometheus alerts
-  Improve testing: unit-testing where possible, e2e and integration testing

pkg/controllers/backup/controller.go
```diff
@@ -209,12 +216,21 @@ func (h *handler) OnBackupChange(_ string, backup *v1.Backup) (*v1.Backup, error
...
 	}
+
 	logrus.Infof("Done with backup")
 	return backup, err
 }
 
 func (h *handler) performBackup(backup *v1.Backup, tmpBackupPath, backupFileName string) error {
 	var err error
+
+	backupStartTS := time.Now()
```
jbiers, line 227: better to move this out of the performBackup function and into onBackupChange

pkg/monitoring/metrics.go
```diff
@@ -0,0 +1,190 @@
...
+			Name: "rancher_backups_failed",
+			Help: "Number of failed Rancher Backups processed by this operator",
+		}, []string{"name"},
+	)
+
+	backupDuration = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "rancher_backup_duration_ms",
+			Help:    "Duration of each backup processed by this operator in ms",
+			Buckets: []float64{500, 750, 1000, 1500, 2000, 2200, 2400, 2600, 2800, 3000, 3200, 3400, 3700, 4000, 4200, 4400, 4800, 5000, 5500, 6000, 7500, 9000, 10000},
```
alexandreLamarre, line 53: IMO this is too granular, for the time being I'd recommend, setting this to something less granular like : `500,1000, 2500, 5000, 7500,,10000`.

---

Metrics like these are useful for customer performance debugging, so the *ideal* bucket boundaries will need to change based on their environments

Ideally we would use exponential histograms (native histograms) to auto-tune bucket boundaries, however this was only introduced in v2.40.X for prometheus and still requires a feature flag on the prometheus side to enable, so we shouldn't have this be enabled by default.

The other option is to have a helm template value that customizes the bucket boundary (probably the best option until expo/native histograms are stabilized)

All that being said having auto/configurable bucket boundaries won't be super useful in the initial metrics introduction
so I'll create a tech debt issue for this

Some further reading for anyone interested:

https://prometheus.io/docs/specs/native_histograms/
https://opentelemetry.io/docs/specs/otel/metrics/data-model/
https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#exponential-histograms

jbiers:
All restore and most backup metrics have passing tests. Two backup metrics are still to have their tests implemented: **rancher_backup_last_processed** and **rancher_backup_duration_ms**. Both have values determined at runtime and not annotated anywhere in the Backup CR, so it's not obvious to me yet how to approach writing tests for them.

mallardduck:
@jbiers - What if we just mock the results on the backup/restore side? That way we can at least create tests to cover the metrics side of this equation. Obviously it'd be most ideal to cover all of the controller code, however covering the metrics side should be better than nothing. Potentially down the road refactors could allow the metrics code in the controller to be more testable.

jbiers:
Followed @mallardduck's idea in having those time-related metrics be tested via unit tests since it's tricky to do via integration tests. (Thanks for the suggestion and the PR :handshake:)

I'm removing the WIP label from the PR and requesting reviews again, as I suppose it is in a good place now with all proposed backup/restore metrics implemented and tested as possible.

pkg/operator/start.go
```diff
@@ -199,18 +217,32 @@ func Run(
 		defaultS3 = s3details
 	}
 
+	if options.shouldRunMetricsServer() {
+		logrus.Info("Starting metrics server")
+		go monitoring.InitMetricsServer(options.MetricsPort)
+
+		logrus.Info("Starting metadata metrics loop")
+		go monitoring.StartBackupMetricsCollection(context.TODO(), c.backupFactory.Resources().V1().Backup(), options.MetricsInterval)
```
mallardduck, line 225: I think this can have the existing `ctx` context passed to it.

pkg/monitoring/metrics.go
```diff
@@ -0,0 +1,232 @@
...
+		}
+
+		updateRestoreMetrics(restoreList.Items)
+	}
+
+	logrus.Info("shutting down restore metrics metadata collection...")
+}
+
+func StartBackupMetricsCollection(
+	_ context.Context,
```
mallardduck, line 186: Oh, or maybe we don't need context at all in here yet and can remove the parameter?

mallardduck (changes requested):
Just a few notes...

pkg/monitoring/metrics.go
jbiers: Yep, we don’t use it! Will be removed ;)

pkg/operator/start.go
```diff
@@ -33,6 +34,9 @@ var (
 
 type RunOptions struct {
 	OperatorPVCEnabled              bool
+	MetricsServerEnabled            bool
+	MetricsPort                     int
+	MetricsInterval                 int
```
alexandreLamarre, line 39: Nit, should be called `MetricsIntervalSeconds` or equivalent, since we have to dig through references to figure out the unit currently

e2e/backup/suite_test.go
```diff
@@ -164,7 +164,10 @@ var _ = BeforeSuite(func() {
 	SetupRancherResourceSet(o)
 
 	errC, ca := SetupOperator(testCtx, restConfig, operator.RunOptions{
-		ChartNamespace: ts.ChartNamespace,
+		ChartNamespace:       ts.ChartNamespace,
+		MetricsServerEnabled: true,
+		MetricsPort:          8080,
+		MetricsInterval:      5,
```
alexandreLamarre, line 170: should be `MetricsInterval:      1`  : speeds up the tests from ~1m -> ~15s. Might not seem important now but as tests are added this could add up

e2e/backup/restore_test.go
alexandreLamarre, line 1: WithTimeout() references should be removed with `MetricInterval = 1` as the global default is 10*time.Second in `suite_test.go` which should be more than enough, could also bump the global default to 15/20 * time.Second if you'd like

e2e/backup/backup_test.go
alexandreLamarre, line 1: WithTimeout() references should be removed with `MetricInterval = 1` as the global default is 10*time.Second in `suite_test.go` which should be more than enough, could also bump the global default to 15/20 * time.Second if you'd like

e2e/backup/backup_test.go
```diff
@@ -60,6 +66,75 @@ const (
 	credentialSecretName = "s3config"
 )
 
+func formatBackupMetrics(backups []string) string {
+	var metrics string
+
+	rancherBackupCountHeader := fmt.Sprintf(`
+	# HELP rancher_backup_count Number of existing Rancher Backup CRs
+	# TYPE rancher_backup_count gauge
+	`)
```
alexandreLamarre, line 75: nit : Some of these fmt.Sprintf(``) are just strings so can use `<string>` instead

e2e/backup/backup_test.go
```diff
@@ -60,6 +66,75 @@ const (
...
+
+	metrics += rancherBackupCountHeader
+	metrics += fmt.Sprintf("rancher_backup_count %d", len(backups))
+
+	rancherBackupsAttemptedHeader := fmt.Sprintf(`
+	# HELP rancher_backups_attempted Number of Rancher Backups processed by this operator
+	# TYPE rancher_backups_attempted counter
+	`)
```
alexandreLamarre, line 83: nit : Some of these fmt.Sprintf(``) in this file are just strings so can use <string> instead

alexandreLamarre (approved):

mallardduck (approved):
LGTM

PR #667, opened 2025-02-14
Add PSA configuration secret to provisioning resource set

pedromfcarvalho:
When a k3s/RKE2 cluster is created with a non-default PSA configuration template, the configuration from that template is then stored in a secret, which is used by the control plane planner.

This secret is only updated by the webhook when the provisioning cluster object is updated, and it wasn't included in the backups created from the default resource set, so after a migration the planner would get stuck in an error state until the secret was re-created manually.

The secret name is defined in the webhook [here](https://github.com/rancher/webhook/blob/v0.6.3/pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go#L39).

This is not needed for RKE1 because in that case the configuration from the template is added to the management v3 cluster object, which is already backed-up, and not a secret.

Issue: https://github.com/rancher/rancher/issues/48884

### DIFF
--- a/charts/rancher-backup/files/default-resourceset-contents/provisioningv2.yaml
+++ b/charts/rancher-backup/files/default-resourceset-contents/provisioningv2.yaml
@@ -13,7 +13,7 @@
   kindsRegexp: "."
 - apiVersion: "v1"
   kindsRegexp: "^secrets$"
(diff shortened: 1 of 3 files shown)


jbiers (dismissed):
LGTM. Just for context, the default resourceSet will stop receiving any updates and be treated as deprecated once https://github.com/rancher/dashboard/issues/12997 is completed. This is part of a larger effort described in https://github.com/rancher/backup-restore-operator/issues/607.

snasovich (approved):

pedromfcarvalho:
Rebased because of a merge conflict.

jbiers (approved):

Issue #654, opened 2025-01-22
[v2.10] rancher-backup start using rancher/kuberlr-kubectl


mallardduck:
See example issue: https://github.com/rancher/backup-restore-operator/issues/652
For example PR and notes about how BRO chart will release for this.

omkarkhatavkar:
### 📝 Rancher Backup & Restore Testing  

**Rancher Version:** `v2.10.4-alpha1`  
**Cluster Version:** `v1.31.5-rc1+rke2r1`  
**Rancher Backup and Restore Version:** `105.0.1+up6.0.1-rc.1`  

#### ✅ Installation Verification  
- Installed the `rancher-backup-restore` chart (`105.0.1+up6.0.1-rc.1`) on Rancher `v2.10.4-alpha1`.  
- Verified that the `rancher-backup-patch-sa` job is using the `rancher/kuberlr-kubectl` image successfully.  
- No errors were encountered during installation.  

![Image](https://github.com/user-attachments/assets/dce15d5d-6d14-4950-959a-393ccf3ac9be)

#### 🔄 Backup and Restore Testing  
- Tested **in-place backup and restore** with both **encrypted and non-encrypted backups**.  
- The backup and restore processes worked as expected. ✅

PR #673, opened 2025-02-27
Use rancher/kuberlr-kubectl image

jbiers:
Issue: https://github.com/rancher/backup-restore-operator/issues/654


mallardduck (changes requested):
This should be targeting [release/v6.x](https://github.com/rancher/backup-restore-operator/tree/release/v6.x) rather than `v6.0`.

PR #688, opened 2025-03-06
ci(gha): fix current build issues with ci

mallardduck:


.github/workflows/ci.yaml
```diff
@@ -60,7 +60,7 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
       - name: Install Helm-unittest
-        run: helm plugin install https://github.com/quintush/helm-unittest
+        run: helm plugin install https://github.com/helm-unittest/helm-unittest
```
mallardduck, line 63: I think this is part of the current failures...not sure why this would affect `hull` CI errors though. @jbiers

PR #742, opened 2025-03-24
[main] Add support for a `FieldSelector` in the ResourceSet

thatmidwesterncoder:
### Problem:
Certain very important secrets are not being backed up by the backup-restore operator when the cluster-name + pool-name gets to be longer than 63 characters, this is because [they fail to match the v2prov regex](https://github.com/rancher/backup-restore-operator/blob/4a35b5ea6009fb099322b709263f4b04b98501a7/charts/rancher-backup/files/default-resourceset-contents/provisioningv2.yaml#L14-L18). 

### Solution:
Add support for selecting resources via `fieldSelector` in order to be able to effectively filter on specific secret types that we care about. This will be useful for other "custom" secret types as well as being able to narrow things that have a certain label + fieldSelector. 

The solution is implemented by just tacking on a fieldSelector to the initial API listing if it is present. By default it selects everything mimicking old behavior (exactly like the label selector api).

### Alternatives:
- I did go down the route of limiting the secret name length [here](https://github.com/rancher/rancher/pull/49322) but that seems to introduce a lot of regression risk as well as feeling really hacky
- Changing rancher to use annotations/labels etc rather than just string names, but once again this would have high regression risk since we would effectively be completely changing how rancher looks up secrets pertaining to a cluster/pool for machine-state/machine-plan/cluster-state secrets

### Testing:
I added e2e integration tests for this which fetch the tarball and extract it into memory - and then validate that the custom-resource-set was used and backed up a single secret (and not the other one)

### DIFF
--- a/charts/rancher-backup/files/default-resourceset-contents/provisioningv2.yaml
+++ b/charts/rancher-backup/files/default-resourceset-contents/provisioningv2.yaml
@@ -13,7 +13,7 @@
   kindsRegexp: "."
 - apiVersion: "v1"
   kindsRegexp: "^secrets$"
-  resourceNameRegexp: "machine-plan$|rke-state$|machine-state$|machine-driver-secret$|machine-provision$|admission-configuration-psact$|^harvesterconfig|^registryconfig-auth|^harvester-cloud-provider-config"
+  resourceNameRegexp: "machine-driver-secret$|machine-provision$|admission-configuration-psact$|^harvesterconfig|^registryconfig-auth|^harvester-cloud-provider-config"
   namespaces:
   - "fleet-default"
 - apiVersion: "v1"
@@ -21,3 +21,21 @@
   resourceNames:
     - "provisioning-log"
   namespaceRegexp: "^c-m-"
+- apiVersion: "v1"
+  kindsRegexp: "^secrets$"
+  namespaces:
+   - "fleet-default"
+  fieldSelectors:
+    "type": "rke.cattle.io/machine-plan"
+- apiVersion: "v1"
+  kindsRegexp: "^secrets$"
+  namespaces:
+   - "fleet-default"
+  fieldSelectors:
+    "type": "rke.cattle.io/cluster-state"
+- apiVersion: "v1"
+  kindsRegexp: "^secrets$"
+  namespaces:
+   - "fleet-default"
+  fieldSelectors:
+    "type": "rke.cattle.io/machine-state"
--- a/charts/rancher-backup/files/sensitive-resourceset-contents/provisioningv2.yaml
+++ b/charts/rancher-backup/files/sensitive-resourceset-contents/provisioningv2.yaml
@@ -1,5 +1,23 @@
 - apiVersion: "v1"
   kindsRegexp: "^secrets$"
-  resourceNameRegexp: "machine-plan$|rke-state$|machine-state$|machine-driver-secret$|machine-provision$|admission-configuration-psact$|^harvesterconfig|^registryconfig-auth|^harvester-cloud-provider-config"
+  resourceNameRegexp: "machine-driver-secret$|machine-provision$|admission-configuration-psact$|^harvesterconfig|^registryconfig-auth|^harvester-cloud-provider-config"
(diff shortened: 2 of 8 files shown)


charts/rancher-backup/files/default-resourceset-contents/provisioningv2.yaml
```diff
@@ -21,3 +21,21 @@
   resourceNames:
     - "provisioning-log"
   namespaceRegexp: "^c-m-"
+- apiVersion: "v1"
+  kindsRegexp: "^secrets$"
+  namespaces:
+   - "fleet-default"
```
jakefhyde, line 27: not a problem for this PR, but what do we do about clusters in a different namespace than fleet-default? cc @rancher/observation-backup

charts/rancher-backup/files/default-resourceset-contents/provisioningv2.yaml
```diff
@@ -21,3 +21,21 @@
   resourceNames:
     - "provisioning-log"
   namespaceRegexp: "^c-m-"
+- apiVersion: "v1"
```
jakefhyde, line 24: Might be a good idea to remove them from line 16.

jakefhyde, line 24: Might be a good idea to remove them from line 16.

charts/rancher-backup/files/sensitive-resourceset-contents/provisioningv2.yaml
```diff
@@ -3,3 +3,21 @@
   resourceNameRegexp: "machine-plan$|rke-state$|machine-state$|machine-driver-secret$|machine-provision$|admission-configuration-psact$|^harvesterconfig|^registryconfig-auth|^harvester-cloud-provider-config"
```
jakefhyde, line 3: Same as previous comment.

jakefhyde (changes requested):

jakefhyde (approved):

mallardduck:
Looks like @jakefhyde gave it an approval so I've assigned @jbiers for O&B side review before merge. Jake, is this something that is expected to backport to 2.11 (or other versions)? If yes, is it considered a critical issue that should be targeting 2.11.1? Forget if we covered that some where else but just want to verify the scope of this for backport considerations.

charts/rancher-backup/files/default-resourceset-contents/provisioningv2.yaml
jbiers: This is indeed a problem and there's an open issue for it here: https://github.com/rancher/backup-restore-operator/issues/574

A workaround is to use the "resources.cattle.io/backup=true" label to make sure the secret is included in the backup.

pkg/resourcesets/collector.go
```diff
@@ -130,12 +131,12 @@ func (h *ResourceHandler) gatherObjectsForResource(ctx context.Context, res k8sv
 
 	// only resources that match name+namespace+label combination will be backed up, so we can filter in any order
```
jbiers, line 132: nit: let's update this comment to better describe the behavior now that we're including field selectors

jbiers (changes requested):
Minor nitpick just to keep consistency in the codebase, once fix will be ready for approval

jbiers (approved):
LGTM

mallardduck:
I think we may need 2 issues to track as with one, only one of the QA teams can mark it as done. I believe when the first team does that it would (likely) be hidden for the following team. Since their filters probably auto mark things as done if the issue is closed too.

jbiers:
Created https://github.com/rancher/rancher/issues/49647 to track the issue on our side

mallardduck:
Hopefully didn't pull the gun too early @thatmidwesterncoder / @jakefhyde ...this came up in our team sync up and with 2.11.1 looming we want to get this to QA ASAP. So I'm also preparing an RC and then making `rancher/charts` PR. Once merged into charts and ready for testing I'll update each respective PR with notes on what version to test with.

PR #754, opened 2025-04-01
[release/v6.x] Add support for a FieldSelector in the ResourceSet

thatmidwesterncoder:
## Issue: https://github.com/rancher/rancher/issues/49661

This is a backport of #742


mallardduck (commented):
Just wondering if we can also include e2e tests from main? If there's too many differences (possible I'm not sure if we've fully backported some CI changes) then that's fine to exclude them and our team will make a tech-debt item to follow up.

Issue #767, opened 2025-05-07
Security best practices


jbiers:
Customers consistently request that our container images and charts are abiding by Kubernetes Best Practices. An example being SURE-7910.

This team-specific issue tracks the components below:

rancher/backup-restore-operator
For the recommendations:

1. Restrict container from acquiring additional privileges (`securityContext.allowPrivilegeEscalation`)
2. Mount container's root filesystem as read only (`securityContext.readOnlyRootFilesystem`)
3. Container is running as root (either via USER on Dockerfile or `securityContext.runAsUser`)

If one of the recommendations above cannot be met, please specify the reason so that we can document and pass that on to customers in future requests.

Reference: SURE-8269

jbiers:
I've opened an initial PR to make sure the BRO chart is compliant with the best practices as best as possible.

Notably, BRO creates files locally while building the backups, so setting **readOnlyRootFilesystem** to true breaks the application and is therefore not possible.
Also, in a specific scenario where the user chooses to store backups in a PV of type HostPath, the process actually has to be running as root as only the root user can write to PVs of this type.

For those reasons I chose to make the **runAsUser** setting optional, having a warning about security in the values file.

It would be good if we could get a code review from the security team on this one.

omkarkhatavkar:
### Tests

| Test ID | Check                                               | Expected                                | Result                 |
| ------- | --------------------------------------------------- | --------------------------------------- | ---------------------- |
| TC-1    | `allowPrivilegeEscalation` is false                 | `false`                                 | ✅ Pass                 |
| TC-2    | Container runs as non-root user                     | `runAsNonRoot: true`, `runAsUser: 1000` | ✅ Pass                 |
| TC-3    | `readOnlyRootFilesystem` is true                    | `true`                                  |  **Not Implemented** (justified in [comment](https://github.com/rancher/backup-restore-operator/issues/767#issuecomment-2884761443)) |
| TC-4    | Pod-level `fsGroup` is set                          | `fsGroup: 1000`                         | ✅ Pass                 |
| TC-5    | SecurityContext is defined at container & pod level | Present                                 | ✅ Pass                 |

### Results 

```
root@ip-10-0-0-25:/home/ubuntu# kubectl get pod rancher-backup-7f6cbf879-8xd2p -n cattle-resources-system -o jsonpath='{.spec.containers[0].securityContext.allowPrivilegeEscalation}'; echo
false
[...]

PR #768, opened 2025-05-13
[main] Make BRO compliant with security best practices

jbiers:
https://github.com/rancher/backup-restore-operator/issues/767

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -12,5 +12,12 @@ RUN ./scripts/build
 
 FROM registry.suse.com/bci/bci-micro:latest
 COPY --from=builder /usr/src/app/bin/backup-restore-operator  /usr/bin/
-USER 1000:1000
+
+RUN echo 'bro:x:1000:1000::/home/bro:/bin/bash' >> /etc/passwd && \
+    echo 'bro:x:1000:' >> /etc/group && \
+    mkdir /home/bro && \
+    chown -R bro:bro /home/bro
+COPY --from=builder /usr/src/app/bin/backup-restore-operator  /usr/bin/
+USER bro
+
 ENTRYPOINT ["backup-restore-operator"]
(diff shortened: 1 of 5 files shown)


jbiers:
I've opened an initial PR to make sure the BRO chart is compliant with the best practices as best as possible.

Notably, BRO creates files locally while building the backups, so setting readOnlyRootFilesystem to true breaks the application and is therefore not possible.
Also, in a specific scenario where the user chooses to store backups in a PV of type HostPath, the process actually has to be running as root as only the root user can write to PVs of this type.

For those reasons I chose to make the runAsUser setting optional, having a warning about security in the values file.

It would be good if we could get a code review from the security team on this one.

andypitcher (dismissed):
Hey @jbiers, as discussed offline the changes LGTM. If possible, Aligning the `Dockerfile` to have the same standard `USER` would be an extra enforcement so that the image always runs as it (independently of the helm chart's `securitycontext`).

jbiers:
> Hey @jbiers, as discussed offline the changes LGTM. If possible, Aligning the `Dockerfile` to have the same standard `USER` would be an extra enforcement so that the image always runs as it (independently of the helm chart's `securitycontext`).

I've added the change requested to the Dockerfile :D 
Thanks for taking the time to look at this

andypitcher (dismissed):

package/Dockerfile
```diff
@@ -12,4 +12,5 @@ RUN ./scripts/build
 
 FROM registry.suse.com/bci/bci-micro:latest
 COPY --from=builder /usr/src/app/bin/backup-restore-operator  /usr/bin/
+USER 1000:1000
```
alexandreLamarre, line 15: Could consider using the pattern in:
https://github.com/rancher/prometheus-federator/blob/44a37ffd9d67a4c9f7922eb64e8bdcfe1f4857fd/package/Dockerfile-helm-locker#L10-L15

Just some small "UX" improvements at the OS layer, and then you can change the constructed object's security context  in `templates` to be *slightly* more readable

alexandreLamarre (dismissed):

mallardduck (approved):

Issue #672, opened 2025-02-25
[RFE] Add Prometheus alerting rules


jbiers:
The Backup&Restore Operator is now able to [export metrics](https://github.com/rancher/backup-restore-operator/pull/648). Alerts should be created to monitor these metrics in a meaningful way.

prachidamle:
SURE-3035

omkarkhatavkar:
@jbiers Can you please add steps how to enable this feature and metrics list.

jbiers:
@omkarkhatavkar For sure!

This PR essentially creates a new alert that will fire when a backup operation fails.
You need to set `.Values.monitoring.metrics.enabled: true`, `.Values.monitoring.serviceMonitor.enabled: true` for the metrics to be exported and `.Values.monitoring.prometheusRules.defaultAlert.enabled: true` for the alert to be created. You should also be able to set custom alerts with the field `.Values.monitoring.prometheusRules.customRules.rules`.

The main thing to be tested here is if the alert is firing correctly on backup failure and if it stops firing once the backup stops failing.

omkarkhatavkar:
### Test Executed :

Rancher Version : 2.12.0-alpha12 

1. Make sure the default rule gets created and alerts should be generated  - **PASSED**

![Image](https://github.com/user-attachments/assets/bc925f3a-1024-4fa5-8fca-3bbb122697a1)

2. Alerts should be removed or deleted if the backups are successful  - **PASSED**

![Image](https://github.com/user-attachments/assets/69c1ab68-a4ff-4915-89ef-977578959e4a)


3. Alerts should be generated for the backup and restore custom rule as well. - **PASSED**

![Image](https://github.com/user-attachments/assets/12ac658e-ceba-46a1-b382-3e108173206e)

PR #769, opened 2025-05-22
[main] Implement alerting

jbiers:
Issue: https://github.com/rancher/backup-restore-operator/issues/672


charts/rancher-backup/values.yaml
```diff
@@ -90,3 +90,19 @@ monitoring:
...
     relabelings: []
+  
+  alerts:
+    enabled: false
+
+    ## Define custom alerting rules here.
+    ## The "BackupFailed" alert is included by default when .Values.monitoring.alerts.enabled is set to true and rancher-monitoring is installed.
+    alertingRules:
+    - alert: BackupFailed
+      expr: rancher_backup{status!~"Completed|"} > 0
```
diogoasouza, line 101: What are the possible values here? There isn't something like `inProgress`?

charts/rancher-backup/templates/alerting-rules.yaml
```diff
@@ -0,0 +1,14 @@
+{{ if and (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") .Values.monitoring.alerts.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: {{ include "backupRestore.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  groups:
+    - name: backup-restore
```
diogoasouza, line 9: No need to support `interval` or `limit`?

charts/rancher-backup/values.yaml
```diff
@@ -90,3 +90,19 @@ monitoring:
...
     metricRelabelings: []
     relabelings: []
+  
+  alerts:
+    enabled: false
+
+    ## Define custom alerting rules here.
+    ## The "BackupFailed" alert is included by default when .Values.monitoring.alerts.enabled is set to true and rancher-monitoring is installed.
+    alertingRules:
+    - alert: BackupFailed
```
mallardduck, line 100: Nice.

Only thing I'm wondering about is, "are there any alerts that BRO should provide by default that cannot be disabled". In other words, are there any features of this new monitoring/alerting area of BRO that (when enabled) are required? I guess this thought goes just as much for these alerting rules as it does for any metrics collection things too - since I could imagine this could relate more to dashboards than alerts TBH. 🤔

charts/rancher-backup/values.yaml
jbiers: Status is empty until it is set to either 'Completed' or to an error message, BRO doesn't set any 'inProgress' equivalents

charts/rancher-backup/values.yaml
```diff
@@ -90,3 +90,19 @@ monitoring:
     additionalLabels: {}
     metricRelabelings: []
     relabelings: []
+  
+  alerts:
+    enabled: false
+
+    ## Define custom alerting rules here.
+    ## The "BackupFailed" alert is included by default when .Values.monitoring.alerts.enabled is set to true and rancher-monitoring is installed.
+    alertingRules:
```
alexandreLamarre, line 99: I would consider renaming this to prometheus rules or something like that to be able to declare both alerting and generic prometheus rules

charts/rancher-backup/values.yaml
alexandreLamarre, line 101: This is fine but  has two issues:
- Fires whenever there is a problem, creating alert fatigue & potential scalability issues
- Fires a unique alert per problematic CR, creating alert fatigue & potential scalability issues

Typically the design of an alerting rule that would catch these errors and reduce alert fatigue so we would do something like:

```yaml
expr :  ((sum(rate(rancher_backups_attempted[5m]))) / (sum(rate(rancher_backups_attempted[5m])) - sum(rate(rancher_backups_failed[5m])))) > 1
```

- This fires when there are persistent backup issues over some window of time
- This aggregates over multiple backup instances

however, we don't have many labels on this metric, so you would also make use of some form of the promQL pattern below: 
```
rancher_backups_failed* on(name) group_left(status) rancher_backup
```

using this directly on the promQL label might be problematic since there is potential for  many-to-many mappings . So to work around that you could considering defining a prometheus rule that creates a new metric with that join when the rule runs:

```yaml
record: rancher_backup_failed_with_status
expr: rancher_backups_failed* on(name) group_left(status) rancher_backup
```
(this many still have many_to_one relationships, so we may need to redesign the rule or metrics slightly)
and use that in place of `rancher_backup_failed` in the above promQL

Disclaimer: I didn't test the promQL above, but it conveys the gist of how we would write the alert with alert fatigue and scale in mind

charts/rancher-backup/values.yaml
```diff
@@ -90,3 +90,19 @@ monitoring:
...
+    - alert: BackupFailed
+      expr: rancher_backup{status!~"Completed|"} > 0
+      for: 1m
+      keep_firing_for: 10m
+      labels:
+          severity: critical
```
alexandreLamarre, line 105: AlertManager uses label matchers to route to specific integrations, so users will want to be able to pass in something along the lines of  `.alerts.default.labels` to add labels in a git-ops friendly way to the labels on the base alerts we provide

alexandreLamarre (changes requested):

charts/rancher-backup/values.yaml
alexandreLamarre, line 99: I would also suggest putting this default alert rule definition in helm template behind a template flag like .Values.useDefaultAlertRules or something and the current definition should be behind `.Values.additionalRules`

charts/rancher-backup/values.yaml
alexandreLamarre: Also should follow naming conventions: https://prometheus.io/docs/practices/rules/

jbiers: As for `expr :  ((sum(rate(rancher_backups_attempted[5m]))) / (sum(rate(rancher_backups_attempted[5m])) - sum(rate(rancher_backups_failed[5m])))) > 1`, I believe it would be `< 1`, correct? Also we need to handle zero division somehow.

Regarding the missing labels, I like the idea of redesigning the _rancher_backups_failed_ and _rancher_backups_attempted_ metrics slightly to also include the status label in them instead of relying on PromQL queries. Do you have anything against that approach?

alexandreLamarre: I'm not opposed so it inherently, but the tradeoff It increases metric cardinality so we potentially want to avoid that.

The above query should be : `>1` since we're measuring `total / (total-failure)` so if we have failures then we get a ratio > 1. We can rework the query to be a ratio between 0 and 1 as well -- but I think exporter libraries typically avoid that in favor of the >1 pattern.


As for zero division you can make us of the or and on operators:
```
(metric_name > 0) * metric_name
  or
on(labels)
1
```

If we're joining on labels

Issue #641, opened 2025-01-07
Remove references to ResourceSet "rancher-resource-set"


jbiers:
The *rancher-resource-set* CR will be deprecated as defined in https://github.com/rancher/backup-restore-operator/issues/607.  It will be kept temporarily in the v2.11 release to give users the chance to migrate their Backups to use the new definitions. This issue tracks the work to be done for complete removal of the *rancher-resource-set* for the v2.12 release.

- [ ] Remove references to it in Docs.
- [ ] Remove it from BRO code.

PR #777, opened 2025-06-05
[main] Remove deprecated resourceset

jbiers:
https://github.com/rancher/backup-restore-operator/issues/641


diogoasouza (commented):
Do we need to remove `charts/rancher-backup-crd/templates/resourceset.yaml`?

jbiers:
@diogoasouza Nope, we're only removing one of the currently existing resourcesets but still deploy other two. So the resourceset CRD definition is still needed :)

PR #785, opened 2025-06-24
[main] Include local project namespaces

jbiers:
https://github.com/rancher/rancher/issues/50638

An issue was identified by @omkarkhatavkar where restores fail if a Project was present in the local cluster.

When creating a Project, a so-called "backing namespace" is automatically created by Rancher with a name in the format `clusterid-projectid`. Those namespaces are included in backups when created in downstream clusters as they match [this rule](https://github.com/rancher/backup-restore-operator/blob/main/charts/rancher-backup/files/basic-resourceset-contents/rancher.yaml#L3) for having a name matching the `^p-` pattern.

However, when a Project is created in a local cluster it ends up missing from the final backups as the `clusterid` is `^local-` instead of `^p-`. As a result, restoring from such a backup errors as the resources living in this namespace can't be created since the namespace itself never comes to exist.


jbiers:
Adding @mallardduck and @alexandreLamarre as reviewers. Do any of you guys see any reasons why those namespaces might have been left out intentionally? If not, then it's a simple fix.

PR #787, opened 2025-06-26
fix: Actually use the default resourcesets for rancher-resource-set-full

viccuad:
The template generation for the `rancher-resource-set-full` was consuming the topic `basic-resourceset-contents` plus the `sensitive-resourceset-contents`, while it should consume the `default-resourceset-contents` plus the `sensitive-resourceset-contents`.

The `default-resourceset-contents` and `sensitive-resourceset-contents` don't have a full intersection, there are Secrets that are covered by one that aren't covered by other.

### DIFF
--- a/charts/rancher-backup/templates/rancher-resourceset-full.yaml
+++ b/charts/rancher-backup/templates/rancher-resourceset-full.yaml
@@ -8,7 +8,7 @@ controllerReferences:
     name: "rancher"
     namespace: "cattle-system"
 resourceSelectors:
-{{- range $path, $_ := .Files.Glob "files/basic-resourceset-contents/*.yaml" -}}
+{{- range $path, $_ := .Files.Glob "files/default-resourceset-contents/*.yaml" -}}
   {{- $.Files.Get $path | nindent 2 -}}
 {{- end -}}
 {{- range $path, $_ := .Files.Glob "files/sensitive-resourceset-contents/*.yaml" -}}


viccuad:
This was discovered in the process of adding Kubewarden to the Backup Operator.

I believe this fixes some errors, such as the first troubleshooting entry listed for Fleet here (yet I haven't tested Fleet backup and restore):

https://ranchermanager.docs.rancher.com/integrations-in-rancher/fleet/overview#troubleshooting

mallardduck (changes requested):
We are actually in the process of removing `default` rules as those are for the old defaults. See @jbiers PR here removing the top level resourceset: https://github.com/rancher/backup-restore-operator/pull/777/files

I believe that instead of doing this to resolve the issues you're seeing, we should instead; a) update `basic` rules to include what is missing, or b) update docs to make it more clear that `rancher-resource-set-full` should be the equivalent of the old default; so comparing what was in `default` to what is in `basic` would be incorrect.

viccuad:
@mallardduck  thanks for the info. Does this removal include the `rancher-resource-set-full` (the template is being touched here)?

Let me summarize to see if I have understood it correctly (from reading 
issue https://github.com/rancher/backup-restore-operator/issues/607)

Currently:
- `basic-resourceset-contents` set. Doesn't contain secrets.
- `default-resourceset-contents` set. Contains secret set A.
- `sensitive-resourceset-contents` set. Contains secret set B, whom may not be a complete superset of A. There may be secrets in A that aren't in B, and viceversa (my hyphothesis on why some Fleet Secrets aren't correctly backed up).
- `rancher-resource-set-full` ResourceSet. Created by joining `basic-resourceset-contents` and `sensitive-resourceset-contents`, as looking at the template it doesn't use `default-resourceset-contents`.

After migration:
- `rancher-resource-set-full` ResourceSet. Created by joining `basic-resourceset-contents` and `sensitive-resourceset-contents`. This is just the same as before migration.

I still think that secrets in `default-resourceset-contents` are not in `sensitive-resourceset-contents`, at least for the Fleet case, but I'm not sure of exactly what is the rationale of the `sensitive-resourceset-contents` set, and I'm not a Fleet developer, so I'm wary of touching the Fleet resource sets.

For Kubewarden, I made sure to have the Secrets also in the `sensitive-resourceset-contents` set, so there will be no problem.

If my explanation makes sense, feel free to close this PR.

viccuad:
Thanks both for the explanations.

I diffed myself, and cannot find the Secrets I thought were missing (did this by checking out https://github.com/rancher/backup-restore-operator/commit/94ff858f6864bc5f9913ec3385af8c4eb78309e2, the parent commit to the [commit](https://github.com/rancher/backup-restore-operator/commit/11976e162a16cf99da0e27ed364fbe2cc199fde8#diff-46304c513c45ee51914fa5469e3765fcb4cd24d7025248e7c1126d8b46ac4ecb that removes the default set, and diffing the basic, default, and sensitive sets for several sets).

I suppose that when building the Kubewarden resource set I made the mistake of using the default set and not the basic one.


Closing then. Will amend my Kubewarden PR to not make use of a now-removed default set.

Issue #786, opened 2025-06-25
Missing Resource Requests & Limits in Rancher Backup-Restore Operator


omkarkhatavkar:
### Summary

The `rancher/backup-restore-operator` pod currently lacks CPU and memory `requests` and `limits`. This can lead to scheduling issues or unexpected evictions under resource pressure.

### Expected

Define default resource limits in the Helm chart:

```yaml
resources:
  requests:
    cpu: 100m
    memory: 256Mi
  limits:
    cpu: 500m
    memory: 1Gi

PR #800, opened 2025-08-05
feat: add extraObjects + resources to the Helm Chart

cest-pas-faux:
Added two small features to the helm chart : 

- extraObjects : to create more objects managed by the release (Backups)
  - we currently have to manage those objects using another system, this will ease the deployment
- resources : to define resources to the rancher-backup-operator container
  - we rely on limitRanges to set default limits, it should be provided by the deployment

Also fixes https://github.com/rancher/backup-restore-operator/issues/786


diogoasouza:
Hi @cest-pas-faux 
Can you keep just the changes to `resources` in this PR? 
We will need to discuss and create a new issue for `extraObjects `, so those changes should be in another PR

cest-pas-faux:
@diogoasouza done

Issue #802, opened 2025-08-12
Unexpected results related to failed Recurring Backup and Metrics


mallardduck:
# Issue:

> The Failed Recurring Backups over time chart on the Rancher monitoring dashboard appears to be non-functional. It seems that when a recurring backup fails, its status changes to pending and the backup does not retry until the issue is resolved. Consequently, the backup does not recur automatically after a failure.

Follow up of the minor issue noted here: https://github.com/rancher/ob-team-charts/issues/107#issuecomment-3178734786

---

Per the reporting, we should investigate the lifecycle of how Recurring backups recover from a failure condition and subsequently how that affects Metrics and what the desired/expected results here should be.

omkarkhatavkar:
### ✅ Rancher Backup and Restore Testing Report  

**Verified On:**  
- **Rancher:** `2.12.2-alpha2`  
- **RKE2 Version:** `v1.33.3+rke2r1`  
- **Backup & Restore Chart Version:** `107.1.0+up8.1.0-rc.1`  

**Observations:**  
1. The **“Failed Recurring Backups over time”** chart is now displayed correctly.  
2. The **Backup Type** is correctly shown as *Recurring* (previously it was not displayed correctly).  

📸 **Screenshot:** 

<img width="823" height="1562" alt="Image" src="https://github.com/user-attachments/assets/1ab835c5-1dd4-49da-b197-e5dac8360eb8" />

PR #803, opened 2025-08-12
[main] Have backupType status update happen earlier in the controller loop

jbiers:
https://github.com/rancher/backup-restore-operator/issues/802

BRO sets the _status.backupType_ field only after fully processing a backup. This causes backups to only get a backupType value if they succeed at least once. As this value is important for metrics filtering in the newly created BRO Grafana dashboards, we should move that to be done before the backup can be error out.

### DIFF
--- a/charts/rancher-backup-crd/templates/backup.yaml
+++ b/charts/rancher-backup-crd/templates/backup.yaml
@@ -44,7 +44,7 @@ spec:
                 nullable: true
                 type: string
               retentionCount:
-                default: 10
+                default: 1
                 minimum: 1
                 type: integer
               schedule:
--- a/pkg/crds/crd.go
+++ b/pkg/crds/crd.go
@@ -100,7 +100,7 @@ func customizeBackup(backup *apiext.CustomResourceDefinition) {
 		retentionCount := spec.Properties["retentionCount"]
 		retentionCount.Nullable = false
 		retentionCount.Default = &apiext.JSON{
-			Raw: []byte("10"),
+			Raw: []byte("1"),
 		}
 		retentionCount.Minimum = &minRetentionCount
 		spec.Properties["retentionCount"] = retentionCount


jbiers:
Added "do not merge" label as this will only go in for v2.12.2.

pkg/controllers/backup/controller.go
```diff
@@ -108,11 +108,14 @@ func (h *handler) OnBackupChange(_ string, backup *v1.Backup) (*v1.Backup, error
 		return h.setReconcilingCondition(backup, err)
```
mallardduck, line 108: I can't do a suggestion here - both because multi-line suggestions seem broken right now, but also because the line I'd want to start at is not in the diff 😿 .

Anyway, I think that doing something like this might make this even cleaner:

```go
validated := backup.DeepCopy()
if err = h.validateBackupSpec(validated); err != nil {
    return h.setReconcilingCondition(backup, err)
}

if !reflect.DeepEqual(backup, validated) {
  if _, err = h.backups.UpdateStatus(backup); err != nil {
      return h.setReconcilingCondition(backup, err)
  }

  return backup, nil
}
```

Because then after this is done once (per backup) then we could actually rely on the `backup.Status.BackupType` to be set (if I'm following correctly). And that could allow us to clean up some logic to use those instead of more "magic" feeling conditions.

pkg/controllers/backup/controller.go
```diff
@@ -108,11 +108,14 @@ func (h *handler) OnBackupChange(_ string, backup *v1.Backup) (*v1.Backup, error
...
 	}
 
+	if _, err = h.backups.UpdateStatus(backup); err != nil {
+		return h.setReconcilingCondition(backup, err)
+	}
+
 	if backup.Status.LastSnapshotTS != "" {
 		if backup.Spec.Schedule == "" { // one-time backup
```
mallardduck, line 116: For instance, if we know that validated backups get saved first, then here the logic can be:
```suggestion
		if backup.Status.BackupType == v1.OneTimeBackupType { // one-time backup
```

Similar to the logic update done in `backupIsSingularAndComplete`

pkg/controllers/backup/controller.go
mallardduck: At least AFAIK, this is how things should be done based on what I learned working with Alex. Since adding the return in after update should enqueue it again (_essentially_) and subsequent runs should be equal and skip the `.UpdateStatus` call all together.

mallardduck (changes requested):

pkg/controllers/backup/controller.go
```diff
@@ -104,15 +105,23 @@ func (h *handler) OnBackupChange(_ string, backup *v1.Backup) (*v1.Backup, error
...
 	logrus.Infof("Processing backup %v", backup.Name)
 
-	if err = h.validateBackupSpec(backup); err != nil {
+	validatedBackup := backup.DeepCopy()
+	if err = h.validateBackupSpec(validatedBackup); err != nil {
 		return h.setReconcilingCondition(backup, err)
 	}
 
+	if !reflect.DeepEqual(backup, validatedBackup) {
+		if validatedBackup, err = h.backups.UpdateStatus(validatedBackup); err != nil {
```
mallardduck, line 114: Looking over the validate function again I just realized that it changes both Spec and Status. So doing `UpdateStatus` alone is not going to cover all potential reasons it's not equal.

mallardduck (changes requested):

pkg/controllers/backup/controller.go
mallardduck: I believe that we need to keep the default to `1` to keep things making sense for `OneTime` backups. Since the default at the CRD level needs to account for both backup types.

We could adjust to be made non-nullable since seeing a retention value of 1 of a `OneTime` type backup could make sense.

I believe that we need to either: a) Adjust how things work to not allow changing the type after a backup is created (breaking change so likely off-limits for now), OR b) ensure the controller can still handle this change (but do it in a more elegant way).

jbiers: Also if you notice the changes applied to the CRD chart, the only change added was really the default value, meaning the field was already handled as non-nullable.

mallardduck: Yep sounds good...this just made me consider something:
> If the backup was created manually and without retentionCount set, the difference would be it's now going from 1 to 10 instead of non-existing-field to 10.

If someone creates a recurring backup outside the UI with no value set, what will the results end up being? I'm wondering if my suggestion causes that to behave different - if so then we may need to keep a check in the controller to them to 10.

In other words, by fixing the "OneTime" to have a more logical default of 1, will that also change "Recurring" to have the same default. Ugh. Sorry my idea might have been a can of worms. 😆

jbiers: Hmmm interesting scenario. And yes, it will for sure affect the default value for Recurring backups created manually :thinking:

PR #812, opened 2025-08-28
 [main] Have backupType status update happen earlier in the controller loop - second approach

jbiers:
https://github.com/rancher/backup-restore-operator/issues/802

BRO sets the status.backupType field only after fully processing a backup. This causes backups to only get a backupType value if they succeed at least once. As this value is important for metrics filtering in the newly created BRO Grafana dashboards, we should move that to be done before the backup can be error out.

This is a second approach to solving the issue.

### DIFF
--- a/pkg/controllers/backup/controller.go
+++ b/pkg/controllers/backup/controller.go
@@ -119,17 +119,8 @@ func (h *handler) OnBackupChange(_ string, backup *v1.Backup) (*v1.Backup, error
 		return backupWithType, nil
 	}
 
-	validatedBackup := backup.DeepCopy()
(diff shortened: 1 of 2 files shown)


jbiers:
> Oh yeah this is a lot more simple feeling than the other approach. I like that because I'm hoping it helps keep things more flexible for later big changes we're hoping to work on.
> 
> Only reason I'm setting to "Request Changes" is I'd like one (or a few) more unit test to cover `setBackupType` and `validateBackupSpec` stuff. Happy to help add that test too btw, just would feel good to have that added too.

Thanks for the thorough review in this and the previous PR. I've added the extra tests. LMK if it looks good to you

mallardduck (approved):
LGTM excellent work!

PR #822, opened 2025-10-07
feat: Add openreports.io CRDs to Kubewarden resourceset

viccuad:
Part of https://github.com/kubewarden/audit-scanner/issues/532, unblocked recently, and which is scheduled to be released in Kubewarden 1.30 at the end of October.

This PR adds the openreports.io CRDs to the Kubewarden resourceset, to be backed up and restored when `.Values.optionalResources.kubewarden.enabled` is `true`.


jbiers (approved):

jbiers:
@viccuad It’s still possible, we just need to release a new rc version to include your new change. It may take a few days as I also want to include another (unrelated) change in the new rc, but it will come out in time for v2.13 :)

PR #908, opened 2026-03-11
[RFE] Add `bro-tool` CLI for support use

mallardduck:
Introduces `bro-tool`, a companion CLI for support engineers and developers working with the Backup and Restore Operator.

The tool operates entirely offline — it fetches or reads a BRO helm chart and inspects its bundled ResourceSet configuration without requiring a live cluster.

## What's included:

  - pkg/version — moved version info to a shared package so both the operator and bro-tool can use it.
  - Multi-module Go workspace (go.work) covering the root operator, cmd/tool, and tests/, with dependency isolation enforced per module.
  - resource-set:view — renders all ResourceSet rules from a given BRO chart version, with table, yaml, and json output. Each rule is attributed back to its source file within the chart (e.g. fleet.yaml, aks.yaml).
  - resource-set:check — checks whether a specific Kubernetes resource would be covered by a backup. Accepts a kind/name string or a full resource manifest, infers API versions for built-in Kubernetes kinds, reports all matching rules with caveats for conditions that couldn't be verified offline, and exits non-zero when no match is found (script-friendly).
  - Chart fetching with local caching (~/.cache/bro-tool/charts/) when using --version.
  - scripts/tidy for one-command tidy+sync across all workspace modules; make build now produces both bin/backup-restore-operator and bin/bro-tool.
  - CI updates: golangci-lint now runs per-module via matrix, and a new workspace-sync job catches go.work.sum drift on every PR.
  - User documentation in docs/bro-tool.md and workspace/contributing notes in DEVELOPING.md.

> [!Note]
>  bro-tool is explicitly out-of-scope for SUSE/Rancher support SLAs and is documented as such.

### DIFF
--- a/cmd/tool/internal/cmd/resourcesetcheck/infer_crd_versions.go
+++ b/cmd/tool/internal/cmd/resourcesetcheck/infer_crd_versions.go
@@ -47,23 +47,24 @@ func buildKindToAPIVersionMap() map[string]string {
 	}
 
 	for gk, gvks := range byGroupKind {
-		// Prefer the scheme's prioritized versions for that API group.
-		for _, gv := range scheme.Scheme.PrioritizedVersionsForGroup(gk.Group) {
-			for _, gvk := range gvks {
-				if gvk.Group == gv.Group && gvk.Version == gv.Version {
-					out[strings.ToLower(gvk.Kind)] = gv.String()
-					goto nextKind
-				}
-			}
-		}
-
-		// Fallback: pick the first external version we saw.
-		out[strings.ToLower(gvks[0].Kind)] = gvks[0].GroupVersion().String()
-
-	nextKind:
+		out[gk.Kind] = preferredAPIVersion(gk.Group, gvks)
 	}
 
 	return out
+}
+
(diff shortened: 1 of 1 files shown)


cmd/tool/internal/chart/render.go
```diff
@@ -0,0 +1,176 @@
...
+			continue
+		}
+		var selectors []v1.ResourceSelector
+		if err := yaml.Unmarshal(f.Data, &selectors); err != nil || len(selectors) == 0 {
+			continue
+		}
+		base := stripFilesPrefix(f.Name)
+		for _, sel := range selectors {
+			if fp, err := selectorFingerprint(sel); err == nil {
+				index[fp] = base
```
susesamu: can we have more than one file with the same selector? if we have files/a.yaml and files/b.yaml, the index would keep just b.yaml, right?

mallardduck: I don't think that should happen in a way where the two files (with the duplciate selectors) should render to the same resource set....but technically it's legal and nothing would prevent that. So you have a solid point and found a gap in how accurate the tool is. So I will fix this.

PR #932, opened 2026-03-31
[ci] Remove hull for chart tests & replace with other tooling

mallardduck:
Per title this change remove the need of `rancher/hull` for chart testing usage in BRO. Instead we more heavily use `helm unittest` and `chart-testing`.

As a stop gap for future renovate improvements I have added local scripts to aid with updating checksums of the CI dependencies. The checksums are found in `.github/workflows/scripts/tool-versions.sh` and the tool to update them in `.github/workflows/scripts/update-checksums.sh`.


susesamu (approved):
LGTM

PR #936, opened 2026-04-07
[Backport v9.x] Remove hull for chart tests & replace with other tooling

mallardduck:
Backports of:
- https://github.com/rancher/backup-restore-operator/pull/932
- https://github.com/rancher/backup-restore-operator/pull/937
- Additional commit to fix branch specific values


diogoasouza (changes requested):
I think you missed updating `scripts/validate` and ended up removing `go.work`

mallardduck:
@diogoasouza - yeah 100%. it's a little odd as we only have workspaces on `main` for the `tests` and `tool`. On older branches we don't have `tool` (yet) so won't need workspaces.

---

Edit: attempted fixes still have some gaps...instead of this I will make a PR to `main` to fix the scripts so that we can try to keep things fairly uniform. I think in the end this will create a cleaner system to maintain. As we can't really control code-drift in the actual operator code, but for CI/Tooling it's more reasonable to keep things similar across branches.

.github/workflows/ci.yaml
```diff
@@ -97,8 +97,8 @@ jobs:
           - x64
           - arm64
         K3S_VERSION:
-          - v1.30.9-k3s1
-          - v1.32.1-k3s1
+          - v1.33.10-k3s1
+          - v1.35.3-k3s1
```
diogoasouza: `release/v9.x` is for Rancher 2.13.x, so this should be v1.32 - v1.34

diogoasouza (changes requested):

.github/workflows/ci.yaml
mallardduck: ty again. fixed now.

diogoasouza (approved):

PR #948, opened 2026-04-16
[docs] Add contributing guide for ResourceSet rules

mallardduck:


rmoioliveira (approved):
Look good to me!

docs/contributing-resourcesets.md
```diff
@@ -0,0 +1,387 @@
...
+
+#### System-charts and the Rancher startup ordering problem
+
+Rancher manages its system charts (Fleet, Webhook, etc.) via Helm, reconciling them when the `rancher-charts` ClusterRepo changes or on a periodic timer (default 6 hours). This creates a layered interaction with BRO restores that feature teams should understand:
+
+- **In-place restore:** BRO scales down the Rancher deployment (driven by the `controllerReferences` field on the ResourceSet — see [What is a ResourceSet?](#what-is-a-resourceset)). BRO restores raw Kubernetes objects from the backup — it does not invoke Helm. BRO also intentionally skips restoring the `rancher` and `rancher-webhook` deployments in `cattle-system` themselves, since those are expected to be reconciled by Helm/Rancher on the way back up rather than overwritten by BRO. Once BRO finishes and scales Rancher back up, Rancher reconciles system charts to match the currently installed Rancher version.
```
anmazzotti, line 380: Just to be explicit on this, I get that system-charts (like Fleet, Turtles) must not be baked up.

For example BRO coverage of Turtles must not include the Turtles controller Deployment, as:
- The Turtles version should not have changed (no version rollback supported)
- Rancher is going to reconcile the Turtles system chart on boot

Did I get it right? Does BRO skip all `cattle-*-system` namespaces by default?

docs/contributing-resourcesets.md
```diff
@@ -0,0 +1,387 @@
...
+> [!NOTE]
+> If this is your first time working with BRO, please read [Background & Reference](#background--reference) first before attempting to write rules.
+
+### Scope and expectations
+
+Before diving in, a few facts that apply to every BRO backup and restore:
+
+- **Local cluster only.** BRO only backs up Kubernetes resources on the local (Rancher management) cluster. Downstream clusters are not backed up — they appear in the backup only as resource objects that describe them in the local cluster. Their actual workloads and state are untouched.
+- **Restores are all-or-nothing.** There is no partial restore. Restoring a backup reverses every change made since the backup was taken, including intentional ones.
+- **Rancher versions must match.** When restoring, the Rancher version on the target cluster must match the version from which the backup was taken. BRO does not handle cross-version restores.
```
anmazzotti, line 14: Is this absolute? Including patch versions like from Rancher v1.2.4 to v1.2.3?

docs/contributing-resourcesets.md
```diff
@@ -0,0 +1,387 @@
...
+
+- **In-place restore:** BRO scales down the Rancher deployment (driven by the `controllerReferences` field on the ResourceSet — see [What is a ResourceSet?](#what-is-a-resourceset)). BRO restores raw Kubernetes objects from the backup — it does not invoke Helm. BRO also intentionally skips restoring the `rancher` and `rancher-webhook` deployments in `cattle-system` themselves, since those are expected to be reconciled by Helm/Rancher on the way back up rather than overwritten by BRO. Once BRO finishes and scales Rancher back up, Rancher reconciles system charts to match the currently installed Rancher version.
+- **Migration (restore to a new cluster):** Rancher is not present at all during the BRO restore phase. The system-chart Helm release objects are restored as raw Kubernetes objects, but Rancher must come up afterwards to drive the charts to a healthy state.
+
+In both cases, the system charts are effectively reconciled by Rancher _after_ BRO completes, not _by_ BRO.
+
+#### CRD API version changes and rollback
```
anmazzotti, line 385: This contradicts the `**Rancher versions must match.**` requirement at the top of this document.
Rollbacks can not be supported.

In terms of cluster provisioning with CAPI, this is officially a not supported scenario.
If someone upgrades let's say the AWS infrastructure provider, and then rollbacks the previous version, even restoring the previous k8s resources state does not guarantee the underlying infrastructure state is going to be the same (and conflicts free).

To reiterate, not even a wipe-and-restore is supported in this scenario.

docs/contributing-resourcesets.md
mallardduck: Ideally yes - the state in the backup was produced by a specific Rancher version. Even Rancher patch differences have potential to cause immediate startup reconciliation. So preventing that as much as possible reduces complexity of debugging situations where things are unhealthy after restore.

docs/contributing-resourcesets.md
mallardduck: I could be reading this wrong but I think the concern you are raising is the same one that is discussed in this long running docs PR: https://github.com/rancher/rancher-docs/pull/1705

At the end of the day the reality is that neither BRO nor `etcd` provide great experiences for users needing to restore after downstream clusters have had intentional changes that they wish to preserve.

docs/contributing-resourcesets.md
mallardduck: > Does BRO skip all cattle-*-system namespaces by default?

BRO backs up nothing by default. Only resources matching rules within the `ResourceSet` selected for the backup will be included; it is a purely opt-in system. This is important because the `ResourceSet` itself is also embedded into the backup file.

At restore time BRO restores everything it finds in the backup - except for 2 very narrow exceptions of "Built-in roles" and the Rancher and webhook deployments. In the case of "in-place" restores the Prune/purge option is used which takes the `ResourceSet` from inside the backup file, then against the live cluster will remove any resources matching it that are not included in the backup.

---

Assuming Turtles resources live in `cattle-turtles-system`, nothing there will be included in a backup until explicit rules are added. They can be put in a new `turtles.yaml` file under `basic-resourceset-contents/` (and `sensitive-resourceset-contents/` for any secrets).

docs/contributing-resourcesets.md
anmazzotti: What is contradictory here is that there is any CRD version bump. 
This is not possible if the Rancher version stays the same, as per requirement.

mallardduck: What do you mean by "Rancher version stays the same"?

A "version rollback" by definition involves 2 different versions of Rancher being involved. One is the "current version" - the one that is running after an update/upgrade - the other is the "backup version" the version that the backup was taken with.

To do a "version rollback" users have 2 options - either attempt in-place restore where BRO scales down Rancher and such, then things are pruned and restored, at the end Rancher should not scale back up. The user will need to either run helm rollback, or helm install using the older version. The other option is to wipe the cluster via re-image - or `rancher-cleanup.sh` - then do a "migration restore" process from the desired backup.

anmazzotti: Line 14 says:
```
**Rancher versions must match.** When restoring, the Rancher version on the target cluster must match the version from which the backup was taken. BRO does not handle cross-version restores.
```

I'm confused by this statement, because to me it means Rancher version must match and BRO should not be used to downgrade a version. I think this can not be supported with CAPI, so it affects Cluster state.
See: https://cluster-api.sigs.k8s.io/reference/versions.html#downgrades

mallardduck: I have updated the doc to provide better clarity around the 2 major factors of a Restore. These are:

```
  Infrastructure axis:
  - In-place — live cluster, Rancher running (BRO pauses it)
  - Migration — fresh/wiped cluster, Rancher absent during restore

  Version axis:
  - Same-version — backup Rancher version = target Rancher version
  - Rollback — backup is older than the currently installed Rancher version
 ```

The reality here is that these have always been supported uses for BRO even if some what imperfect. That was why I linked to the unmerged docs PR as it highlights how imperfect all our solutions to bacup+restore are when it comes to state of downstream clusters.

Comment (mallardduck) [reply docs/contributing-resourcesets.md:385]: > When restoring, the Rancher version on the target cluster must match the version from which the backup was taken. 

This means: if you updated to `2.12.6` but want to restore a backup that was taken on 2.12.4 then upon rancher start up you should use `2.12.4`. This does not mean if you are on `2.12.6` all your backups from prior versions are invalid.

> BRO does not handle cross-version restores.

This means that if you are on Rancher 2.13 and restore a 2.12 backup it will be doing an upgrade upon Rancher start up. This is a "cross-version restore" to make it not a cross-version restore should start 2.12 Rancher as the backup was taken with.

Comment (mallardduck) [reply docs/contributing-resourcesets.md:385]: Added a section to define and explain what "cross-version restores" means.

Comment (anmazzotti) [reply docs/contributing-resourcesets.md:385]: I see what you mean now, thank you for elaborating.
I'm still concerned by the rollback scenario. You say it's imperfect, but factually it's unsupported.

Also the current README states the first use case as:
```
Performing a backup before upgrading Rancher and restoring after a failed upgrade.
```

This in particular we can not guarantee. Updating from Rancher v2.13 to v2.14 for example bring a new CAPI version and API bump, plus all the infrastructure providers. I think this should be stated explicitly that this is not a rollback tool.

I do agree a best effort support is nice to have, just worried about giving the users the wrong impression. I'll resolve this conversation anyway, I think any further discussion is out of scope for this document.

rmoioliveira (approved):

anmazzotti (approved):
Thank you for the great write up!

PR #1008, opened 2026-07-02
[main] Fix Metrics issue causing BRO failing to startup

mallardduck:
This PR:
- Updates BRO CRDs from using obsolete methods to use `controller-gen`
- Adds necessary `kubebuilder` annotations to get equivalent CRD yaml generated
- Fixes some issues with the previous OpenAPI results being generated (most things were `nullable` even if `required`)
- **Fixes a bug** related to metrics which causes BRO to fail to start


gotools/controller-gen/go.mod
```diff
@@ -0,0 +1,62 @@
+module github.com/rancher/rancher/gotools/controller-gen
```
alemorcuq: Is this the expected module?

pkg/apis/resources.cattle.io/v1/common.go
```diff
@@ -0,0 +1,43 @@
...
+	// TODO: Add a setting for varying trace options that minio provides
+	// +optional
+	// +nullable
+	Aws *AwsConfig `json:"aws,omitempty"`
+}
+
+type AwsConfig struct {
+	// +default:value=true
+	// +optional
+	DualStack bool `json:"dualStack,omitempty"`
```
alemorcuq: We need this to be a pointer, otherwise it's never going to be `false`. `omitempty` will read false as empty, and the `+default:value` would then set this to true.

gotools/controller-gen/go.mod
mallardduck: I guess technically not - but also it doesn't really matter in the end. These files come from the `r/r` repo which is where the name comes from. So we could update it but also it's not going to cause any issues really.

pkg/apis/resources.cattle.io/v1/common.go
mallardduck: Great catch - I'm going to revert back to remove `omitempty` and `+optional`. Using pointers here will mean extra checks later and AFAIU the previous annotation worked.

rmoioliveira (dismissed):

alemorcuq (dismissed):

mallardduck:
## What Changed

We migrated CRD generation from **Wrangler v3.7.0** to **controller-gen v0.21.0** (the Kubernetes standard).

## Why the CRDs Look So Different

**99% of the diff is fixing a Wrangler bug.** Wrangler incorrectly added `nullable: true` to ~70 fields that weren't actually nullable (non-pointer types).

**Example:**
```yaml
# Old (Wrangler) - INCORRECT
resourceSetName:
  type: string
  nullable: true      # ❌ Wrong - string isn't a pointer

# New (controller-gen) - CORRECT  
resourceSetName:
  type: string        # ✅ Correct - not nullable
```

The other 1% are improvements:
- Added type formats (`int64`, `int32`)
- Added enum validation (BackupType: `One-time` | `Recurring`)
- Added Kubernetes list type metadata
- Better API documentation

alemorcuq (approved):

rmoioliveira (approved):

susesamu (approved):
LGTM! YOU'RE A WIZARD!

Issue #1011, opened 2026-07-02
[tech-debt] Get rid of `init()` in Monitoring package


mallardduck:
Per title, lets find a better way to handle the init of the metrics related code. Ideally we should be able to remove the `init()` and instead handle the init of those stuff in a more conditional way that only affects BRO when metrics are enabled.

https://github.com/rancher/backup-restore-operator/blob/ae6611373527517afac9edb62f1cc2e26c797fcc/pkg/monitoring/metrics.go#L78-L92

---

Other potential points of concern in Monitoring package:
- Issue #1: Unconditional init() Function (lines 78-102)
- ssue #2: Potential bug on Line 169

mallardduck:
This is in v10.0.8-rc.3

omkarkhatavkar:
@mallardduck Verified this here with test case mentioend https://github.com/rancher/observability-e2e/pull/179 for 2.15.0-alpha21, Test is passed. As I believe this is more like an integrational test of monitoting metrics wtih BRO chart and migration.

mallardduck:
Yeah that is great work @omkarkhatavkar - those tests look great.

PR #1012, opened 2026-07-06
refactor: initialize metrics conditionally

alemorcuq:
Move the unconditional initialization of rancherBackupDurationBuckets into an explicit SetupBackupDurationMetric() that can be conditionally called, ensuring it only runs one time.

It also adds a fallback mechanism to default buckets when there's an error parsing user provided ones.

Issue: #1011

### DIFF
--- a/pkg/monitoring/metrics.go
+++ b/pkg/monitoring/metrics.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


mallardduck:
LGTM on initial review we'll want to merge this in after: https://github.com/rancher/backup-restore-operator/pull/1008
Then I can backport both change-sets at the same time to 10x branch - my PR will need minor downgrades of certain `go.mod` to fit the backport.

rmoioliveira (approved):

mallardduck (approved):
LGTM nicely done

PR #1045, opened 2026-07-24
Automation core

diogoasouza:
https://github.com/rancher/rancher/issues/56201


scripts/ci
mallardduck, line 1: Centralizing dev scripts like these is interesting...my only concern is that these not being local to the code will make DX harder. On one hand making tooling uniform is good, but it should still be as simple as it is today to use. I suspect we can use git worktrees and hooks to make that easy. Or maybe you have other idea for this and they sync from here to release branches?

mallardduck (approved):

PR #1059, opened 2026-08-03
Fix automation-core to work with GitHub Actions security boundaries

mallardduck:


actions/release/action.yaml
```diff
@@ -44,59 +73,51 @@ runs:
...
-      with:
-        secrets: |
-          secret/data/github/repo/${{ github.repository }}/rancher-prime-registry/credentials registry | PRIME_REGISTRY ;
-          secret/data/github/repo/${{ github.repository }}/rancher-prime-registry/credentials username | PRIME_REGISTRY_USERNAME ;
-          secret/data/github/repo/${{ github.repository }}/rancher-prime-registry/credentials password | PRIME_REGISTRY_PASSWORD ;
+      run: |
+        # NOTE: This script is in the automation-core branch at .github/scripts/check-semver.sh
+        # For now, we expect it to exist in the calling repo's .github/scripts/ directory
+        # TODO: In P2, create a check-semver action that wraps this script
+        bash ./.github/scripts/check-semver "${{ github.ref_name }}" >> "$GITHUB_OUTPUT"
```
diogoasouza, line 107: The script is not that big, can't we just use it here directly instead of creating a new file?

diogoasouza (approved):

actions/release/action.yaml
mallardduck: For [phase 3 of the plan](https://github.com/rancher/backup-restore-operator/issues/1062) the action would help each "release branch local" workflow call the same check-semver tooling.

PR #1063, opened 2026-08-03
P2: Create 4 focused atomic actions

mallardduck:
Replace monolithic release action with focused, composable actions:

   Actions created:
   - check-semver: Parse semantic version characteristics (wraps script)
   - compute-branch-tags: Calculate branch tags for head builds (wraps script)
   - undraft-release: Undraft GitHub release with retry logic
   - run-goreleaser: Execute goreleaser and validate outputs

   Workflow updates:
   - ci.yaml: Add go-version input, parameterize container images
   - head-builds.yaml: Use compute-branch-tags action

   Documentation:
   - Updated README.md with full action documentation and examples
   - Documented security boundaries (vault, SLSA attestation)

   Deleted:
   - actions/release (monolithic action from P1)


.github/workflows/ci.yaml
```diff
@@ -10,12 +10,16 @@ on:
...
+        description: Go version to use
+        type: string
+        required: true
 jobs:
   golangci:
     name: lint (${{ matrix.module }})
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/rancher/ci-image/go1.26
+      image: ghcr.io/rancher/ci-image/go${{ inputs.go-version }}
```
diogoasouza: Can't we just keep go 1.26 (or the latest) for this? Go is supposedly backwards compatible

actions/check-semver/action.yaml
```diff
@@ -0,0 +1,24 @@
...
+    description: Whether the tag has build metadata
+    value: ${{ steps.check.outputs.HAS_BUILD_META }}
+
+runs:
+  using: composite
+  steps:
+    - name: Check semver characteristics
+      id: check
+      shell: bash
+      run: bash ${{ github.action_path }}/../../.github/scripts/check-semver.sh "${{ inputs.tag }}" >> "$GITHUB_OUTPUT"
```
diogoasouza: Does this still require the script to be in each release branch?

mallardduck: Nope - just forgot to remove them from the other PR and this was a literal move. I've updated so each script is just a sibling of their action so things are way more clear now.

.github/workflows/ci.yaml
mallardduck: > Go is supposedly backwards compatible

That's fair I guess - ~maybe we make it an optional with default to `1.26`~. My gut just wants this pluggable that way if/when we do have versions we don't want to change, or a new one we want to adopt future go faster we have that option.

🤔 On second thought any default would mean an implicit upgrade if we change the default. So i'd rather not set a default at all. So I think the options are to set all to 1.26 and consider adding this input back if/when needed. Or add it now as part of the improvements before the next phase.

diogoasouza (approved):

PR #1076, opened 2026-08-06
Fix checkout bug

mallardduck:


.github/workflows/ci.yaml
```diff
@@ -31,7 +31,9 @@ jobs:
       - name: Checkout bin-deps config to local branch
         if: steps.check.outputs.exists == 'true'
         shell: bash
-        run: git checkout automation-core -- .bin-deps.yaml
+        run: |
+          git fetch origin automation-core
```
rmoioliveira, line 35: I think that this fetch is not needed because the default `fetch-depth` is already 0.

https://github.com/rancher/backup-restore-operator/actions/runs/31115060773/job/92662334187#step:4:3

rmoioliveira (approved):

diogoasouza (approved):

PR #1194, opened 2026-09-11
Update chart annotations/versions for 2.16

mallardduck:


charts/rancher-backup/Chart.yaml
```diff
@@ -2,12 +2,12 @@ annotations:
...
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: Rancher Backups
-  catalog.cattle.io/kube-version: '>= 1.34.0-0 < 1.37.0-0'
+  catalog.cattle.io/kube-version: '>= 1.35.0-0'
   catalog.cattle.io/namespace: cattle-resources-system
   catalog.cattle.io/os: linux
   catalog.cattle.io/permits-os: linux,windows
   catalog.cattle.io/provides-gvr: resources.cattle.io.resourceset/v1
-  catalog.cattle.io/rancher-version: '>= 2.15.0-0 < 2.16.0-0'
+  catalog.cattle.io/rancher-version: '>= 2.15.0-0'
```
diogoasouza: shouldn't this be 2.16?

alemorcuq: Also, why the removal of the upper part of the interval?

mallardduck: > why the removal of the upper part of the interval?

Great question - essentially this removes a step from the current "charts forwardport" process that happens. And prepares the be ready in advanced for future OCI based distribution where the annotations will have to only set lower limits not upper limits.

diogoasouza (approved):

